PR #109, opened 2022-01-13
Added option to set indentation when writing json file. Also added a null check to remove_excess_players.

JanEricNitschke:
Added the options to set the indentation in json.dump to write_json and also in clean_rounds which calls write_json. Kept the default at None which is the current behavior of write_json (since recently).

Interestingly the json written by parse_demo (where the writing occurs in the go code) seems to employ some indentation making the current defaults a bit inconsistent.

---

Additionally i added a None check to remove_excess_players in clean_rounds. It check whether both sides even have a none None ["players"] entry before cutting on their lengths. If either side has a None entry it accepts the round and gives a warning about that fact.

It would probably be a bit better to check each side one by one so that if one side is None and the other 5 the round is accepted by if the other side is >5 the round is not. Alternatively it could be an option to chose whether rounds with None sides should be kept or removed and if then the default should be set to keep or remove.

### DIFF
--- a/csgo/parser/demoparser.py
+++ b/csgo/parser/demoparser.py
@@ -35,6 +35,7 @@ class DemoParser:
         trade_time=5,
         dmg_rolled=False,
         buy_style="hltv",
+        json_indentation=False,
     ):
         # Set up logger
         if log:
@@ -122,8 +123,10 @@ class DemoParser:
 
         self.dmg_rolled = dmg_rolled
         self.parse_frames = parse_frames
+        self.json_indentation = json_indentation
         self.logger.info("Rollup damages set to " + str(self.dmg_rolled))
         self.logger.info("Parse frames set to " + str(self.parse_frames))
+        self.logger.info("Output json indentation set to " + str(self.json_indentation))
 
         # Set parse error to False
         self.parse_error = False
@@ -179,6 +182,8 @@ class DemoParser:
             self.parser_cmd.append("--dmgrolled")
         if self.parse_frames:
             self.parser_cmd.append("--parseframes")
+        if self.json_indentation:
+            self.parser_cmd.append("--jsonindentation")
         proc = subprocess.Popen(
             self.parser_cmd,
             stdout=subprocess.PIPE,
@@ -605,7 +610,6 @@ class DemoParser:
         remove_excess_kills=True,
         remove_bad_endings=True,
         return_type="json",
-        json_indent=None,
     ):
         """Cleans a parsed demofile JSON.
 
@@ -646,7 +650,7 @@ class DemoParser:
                 self.remove_end_round()
             self.renumber_rounds()
             self.rescore_rounds()
-            self.write_json(indent=json_indent)
+            self.write_json()
             if return_type == "json":
                 return self.json
             elif return_type == "df":
@@ -661,10 +665,10 @@ class DemoParser:
                 "JSON not found. Run .parse() or .read_json() if JSON already exists"
             )
 
-    def write_json(self, indent=None):
+    def write_json(self):
         """Rewrite the JSON file"""
         with open(self.output_file, "w", encoding="utf8") as fp:
-            json.dump(self.json, fp, indent=indent)
+            json.dump(self.json, fp, indent=(1 if self.json_indentation else None))
 
     def renumber_rounds(self):
         """Renumbers the rounds.
--- a/csgo/parser/demoparser.py
+++ b/csgo/parser/demoparser.py
@@ -761,11 +761,14 @@ class DemoParser:
             for r in self.json["gameRounds"]:
                 if len(r["frames"]) > 0:
                     f = r["frames"][0]
-                    if f["ct"]["players"] == None or f["t"]["players"] == None:
-                        cleaned_rounds.append(r)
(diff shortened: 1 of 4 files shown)


JanEricNitschke:
Now committed a change to consider each side individually (and fail the round if both sides have None).
Also removed the warning message since i think this should be the intended behavior for POV demos (that always only have one side).

pnxenopoulos:
Hey Jan-Eric, thanks for bringing this PR up, it's addressing an important issue. Let me try and understand some things, first.

It seems like the main problem we are trying to address here in the `remove_excess_players` function is what to do when a round's first frame returns as having more than 5 people. It seems like the issue you raise is directed at POV demos where we will not have the information from the other team (I'm guessing). I haven't done too much testing with POV demos, do you have one you could link to?

Perhaps there should be a flag for a POV demo specified when one makes the parser. Then, that can be used to call a different cleaning function instead of `remove_excess_players`. What do you think?

Aside from the conversation above, I have a few suggestions:

- I think indentation should be a boolean class parameter that we initially put as `False`, and then a user can specify manually later on, e.g.
```
p = DemoParser(...)
p.indentation = True 
```
which would then indent the JSON by a value of `4`, which I think was the original. What we can then do is pass the `indentation` parameter to the golang parser. What do you think?

- The code here:
```
if f["ct"]["players"] == None:
    if f["t"]["players"] == None:
        pass
    elif len(f["t"]["players"]) <= 5:
        cleaned_rounds.append(r)
elif len(f["ct"]["players"]) <= 5:
    if (f["t"]["players"] == None) or (len(f["t"]["players"]) <= 5):
        cleaned_rounds.append(r)
```
I believe can be condensed down into one if-else, since the outcome of `len(f["ct"]["players"]) <= 5` + `(f["t"]["players"] == None) or (len(f["t"]["players"]) <= 5)` is the same as `f["ct"]["players"] == None` + `len(f["t"]["players"]) <= 5`
- Also, could you write a test for your new addition?

JanEricNitschke:
Yeah, what we want is to remove rounds with more than 5 players on either side. Problem is that it crashes when you have `"players": null` on one side. For the most recent version of the parser there is so far only one case where i encountered that (though there could potentially be others i guess). That one situation is POV demos where one players records the game instead of the server itself. 

I have one as well as the resulting json (after the changes in this PR) here https://cernbox.cern.ch/index.php/s/ZY1Y2jD8Z0nN3aS.

---

```
p = DemoParser(...)
p.indentation = True 
```
I had thought about doing it that way, not sure anymore why i decided to go with the version in this PR. Considering that that allows for better consistency with the go parser it probably makes sense.

---
```
if f["ct"]["players"] == None:
    if f["t"]["players"] == None:
        pass
    elif len(f["t"]["players"]) <= 5:
        cleaned_rounds.append(r)
elif len(f["ct"]["players"]) <= 5:
    if (f["t"]["players"] == None) or (len(f["t"]["players"]) <= 5):
        cleaned_rounds.append(r)
```

If there were no nones in this one could easily do it like:


```
if  not ((f["ct"]["players"] == [] and f["t"]["players"]==[]) or (len(f["t"]["players"]) > 5) or (len(f["t"]["players"]) > 5)):
     cleaned_rounds.append(r)
```

But since in the relevant case the list of players is not empty but None this setup would give you an error if only one side experiences it. Cause then  the first part does not instantly throw the result but goes to the second part when one of them gives the error that None does not have a length.

It might be possible to condense it, but i am not sure tbh.

JanEricNitschke:
Moved the json indentation setting to a class parameter and also propagated it through to the go code.
I have now set it so that the indentation is set to "1" if turned on because that is what was used in the go code.
Alternatively we can change it to "4" in the python part and "    " in the go part. But i think 1 is probably the best option cause it already gives basically all of the readability improvements while having the minimal increase in space.

---

>Perhaps there should be a flag for a POV demo specified when one makes the parser. Then, that can be used to call a different cleaning function instead of remove_excess_players. What do you think?

this can probably be read out of the parsed json if we want to go that route. Normally the "clientName" is set to "GOTV Demo" but POV demos have it set to the player name. (i guess looking at this would cause problems for a player called GOTV Demo)

But i do not really think we need to make this distinction tbh. I think the code i have above work (and all the other code as well) works as one would wish for both general GOTV and POV demos.

JanEricNitschke:
Something like that for the test?
```
    def test_player_clean(self):
        """Tests that remove excess players is working."""
        self.player_clean_parser = DemoParser(
            demofile="41.dem", log=False, parse_frames=True
        )
        self.player_clean_data = self.round_clean_parser.parse()
        self.player_clean_parser.remove_excess_players()
        assert len(self.player_clean_data["gameRounds"]) == 28
```

pnxenopoulos:
> 

Yes that works. Send me `41.dem` and I will upload to the cloud repository.

Then, edit `test_data.json` in the `tests/` directory to have a new entry like:

```
"pov-clean": {
        "url": "https://storage.googleapis.com/csgo-tests/41.dem"
    }
```

pnxenopoulos:
Agreed on your discretion for the indentation, your code looks great (and passes tests). Do you think we can make a test for the indentation functionality? I'm actually not sure how of the top of my head.

Issue #123, opened 2022-01-21
Add parameter to `clean_rounds` to skip saving to JSON


adbok001:
`clean_rounds` saves the cleaned demo data as JSON file. It would be useful to be able to skip the saving process

pnxenopoulos:
Added a `save_to_json` param that you can pass in `clean_rounds(...)`. Defaults to `True`. #128

Issue #122, opened 2022-01-20
Add `velocity` values to playerFrames


adbok001:
Usually, the velocity can be calculated from the change of x, y and z values of a player between two frames. But some demos seem to contain frames, where the movement of two frames is stacked together in one frame, like this:

```
tick                                                             116863
x                                                          -1548.832031
y                                                            115.247444
z                                                             101.39814

tick                                                             116865
x                                                          -1548.832031
y                                                            115.247444
z                                                             101.39814

tick                                                             116867
x                                                          -1557.205933
y                                                            147.287308
z                                                             59.259563
```

This leads to velocity values above the maximum on official servers. There is a [`Velocity()`](https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v2/pkg/demoinfocs/common#Player.Velocity) function in the golang parser, which provides the x, y and z velocity values. They could be added to playerFrames - hopefully they represent the actual velocity in these "laggy" frames.

The velocity shown with `cl_showpos 1` can then be calculated with math.sqrt(vel.x * vel.x + vel.y * vel.y) * tick rate of the demo (with `parse_rate=1`). The actual speed would include the z axis movement as well.

pnxenopoulos:
Added in the `velocityX`, `velocityY` and `velocityZ` flags for player frames. #128

PR #124, opened 2022-01-22
Added velocity values to player frames and added param to skip JSON saving in clean_rounds()

adbok001:
Implementation for #122 and #123 

There are older commits, but I have no idea how to exclude them from the pull request


pnxenopoulos:
Oops -- I forgot this while addressing your issues. Gonna close since I added most of what is in here, DM if any errors.

PR #154, opened 2022-04-25
Fixed a typo

MaukWM:
dealnig -> dealing


pnxenopoulos:
Added this in #157, thanks for pointing it out!

Issue #158, opened 2022-05-02
Division by 0


nicodemuz-zz:
**Describe the bug**
If a player haven't died or haven't killed anyone there is a couple of cases with division by 0 for me in stats.py at lines 204 and 221. I fixed it for myself by adding an if at the end of the rows like this:

- 204: player_statistics[player]["kills"] / player_statistics[player]["deaths"] if player_statistics[player]["deaths"] != 0 else player_statistics[player]["kills"],
- 221: player_statistics[player]["hs"] / player_statistics[player]["kills"] if player_statistics[player]["kills"] != 0 else 0, 2

**To Reproduce**
1. Record a session where the gamer doesn't die or doesn't kill (often only happens with very few rounds recorded)
2. Analyze the recorded .dem
demo_parser = DemoParser(...)
data = demo_parser.parse()
3. See error

**Expected behavior**
Not a division by 0 error, either nothing is calculated or a proper value is chosen based on what is looked for like in the example solution

pnxenopoulos:
Hi @nicodemuz-zz, thanks for spotting and reporting this error! 

How crucial is this error to your workflow? If it isn't crucial, I'll just wrap it up in the next release. If it's important, I can make a separate release just for this.

nicodemuz-zz:
Not crucial at all, so wrap it up for a future release. :)

pnxenopoulos:
Keeping open so I don't forget!

pnxenopoulos:
Fixed in #160

Issue #152, opened 2022-04-18
Problem with tRoundStartMoney and ctRoundStartMoney (first round of each side)


kenmareirl:
**Describe the bug**
Incorrect amounts for tRoundStartMoney and ctRoundStartMoney but **only** for the first round of each side (other rounds are OK).
Using parse function + clean_rounds (same problem using or not the argument remove_bad_scoring=True).

**To Reproduce**
1. Parse the following demos : [AST-MIBR](https://www.hltv.org/matches/2354338/astralis-vs-mibr-iem-katowice-2022-play-in), [BIG-GOD](https://www.hltv.org/matches/2354569/big-vs-godsent-esl-pro-league-season-15), [G2-VIT](https://www.hltv.org/matches/2353988/g2-vs-vitality-blast-premier-spring-groups-2022)
2. Check the following fields for the JSON : tRoundStartMoney and ctRoundStartMoney 
3. Problems detected: 

- AST-MIBR and BIG-GOD : wrong amounts for both team at round 1 and round 16
- G2-VIT : inferno : wrong amounts for both team at round 16, mirage : wrong amount at round 16 for Team Vitality, incorrect amounts for both team at round 16

**Expected behavior**
For round 1 and 16 the tRoundStartMoney and ctRoundStartMoney should be 4000$ (800$ per player)

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [N/A]
 - Version [N/A]

**Smartphone (please complete the following information):**
 - Device: [N/A]
 - OS: [N/A]
 - Browser [N/A]
 - Version [N/A]

**Additional context**
Other rounds work fine, no problem detected.

pnxenopoulos:
Interesting...when you look at the frames, the money looks fine. I imagine it has to be the fact that I'm grabbing round start money _during_ round start. 

How important is this feature? We could remove it since it is unstable.

pnxenopoulos:
I think I'll have better success with manually implementing these keys, which is what I'll do. Even in frames, I'll just loop through and sum

pnxenopoulos:
Think the best approach here is to just calculate this by looking on the FreezeTimeEnd. Sometimes demos may miss that, not much we can do!

pnxenopoulos:
Addressed in #166

PR #160, opened 2022-05-03
Player side, round scoring, div by 0

pnxenopoulos:
- Fixing player side to update on each frame (reborn in Discord)
- Making `remove_bad_scoring=True` by default
- Fixing issue in #158


pnxenopoulos:
Removing begin money #152

PR #198, opened 2022-11-05
Update nav.py

davidgfb:


JanEricNitschke:
Here i am unsure if this is needed. I definitely feel that it is significantly less readable then the current version.

Especially with the proposed variable names.

pnxenopoulos:
See #208

Issue #200, opened 2022-11-22
Adding 2d option for find_closest_area function


bahtman:
**Is your feature request related to a problem? Please describe.**
Currently building a dashboard that takes user input to define player locations. The input is taken from a 2d map and therefore doesn't have z-axis information. This makes it impossible to select some areas. 

**Describe the solution you'd like**
Add an option to ignore the z-axis. Therefore only computing the euclidian distance wrt. x and y.

JanEricNitschke:
I do not really think that that is possible to be honest.

How would you resolve the ambiguity on maps like nuke? It is probably best to leave this up to the user how they want to handle it.

On flat maps you could just use any z-value and should get the correct result and on multi-level maps you would have to decide this for yourself.

bahtman:
Currently if I pick z=0 on dust2 all parts of lower dark would not be returned, since they are further away from the top of stairs. 

I understand very vertical maps wouldn't be aligned with the solution i propose.

JanEricNitschke:
Ah, yeah. If the horizontal distance is smaller than the horizontal one it returns the wrong value.

This is a pretty simple change. If you can you could just make a PR for this. Otherwise I'll make one tomorrow.

PR #201, opened 2022-11-23
Added 2d parameter to find_closest_area func.

bahtman:


JanEricNitschke:
In response to this issue: https://github.com/pnxenopoulos/awpy/issues/200

JanEricNitschke:
Looks good to me. It would be awesome if you could also add some test coverage for the new option.

JanEricNitschke:
This PR should also be ready now.

Issue #183, opened 2022-08-10
Same nicknames


Iris0o:
**Describe the bug**
If the player changed his nickname in the game, then there will be more elements in player_stats

**Screenshots**
![image](https://user-images.githubusercontent.com/57287423/183906974-5126553b-6d86-4db5-bdc4-23a578ed1eac.png)

[match730_003564561176207556835_0769000825_349 .json.txt](https://github.com/pnxenopoulos/awpy/files/9300195/match730_003564561176207556835_0769000825_349.json.txt)

Iris0o:
I changed my nickname(IRIS0) to my friend's nickname(leshasharf) when he flew out

JanEricNitschke:
Would be solved by changing https://github.com/pnxenopoulos/awpy/blob/main/awpy/analytics/stats.py#L17

`player_statistics[str(p["steamID"]) + " - " + p["playerName"]]` in https://github.com/pnxenopoulos/awpy/blob/main/awpy/analytics/stats.py#L17 (and everywhere else there)
to:
`player_statistics[p["playerName"] if p["steamID"] == 0 else str(p["steamID"])]`

Issue #182, opened 2022-08-07
Stat enhancements


paulsnoops:
It would be nice to be able to see a player's stats as follows:

- Amount of rounds with 0 kill, 1 kill, 2 kills, 3 kills, 4 kills, 5 kills
- Clutches: amount of rounds where a player had a clutch for each of 1v1, 1v2, 1v3, 1v4, 1v5 and how successful they were; as an example player A had 4 rounds where they were 1v3 and won 50% of those clutches (another example in pic 1 below)
- I think we can already see how many trade kills a player had but what about how many times they were traded by another player?
- Player saved by teammates/Player saved teammates, both the amounts and per round (as seen in picture 2 like on HLTV)

**Pic 1**
![image](https://user-images.githubusercontent.com/18243520/183278891-5d329004-cb0a-467d-b1d2-35a0ac2b17d9.png)

**Pic 2**
![image](https://user-images.githubusercontent.com/18243520/183278832-42f4493e-f78e-431b-bc48-f0d1c4310cb9.png)

I realise some of this may be difficult to implement so no worries if they aren't added :)
Thank you

JanEricNitschke:
I think multi-kill rounds are probably easy to track and add to the player statistics function.

Clutches are probably also possible but slightly harder.

I think traded is actually trivial since it is already effectively being monitored for KAST.

I dont know about saving/being saved. There might actually be an event in the demo file since this is broadcasted in chat in MM right? If not then one would need to chose how to define that.

If you are interested you could probably try to implement some of these here: https://github.com/pnxenopoulos/awpy/blob/main/awpy/analytics/stats.py

JanEricNitschke:
Most of these things have been implemented with https://github.com/pnxenopoulos/awpy/commit/142583e72d2de9990a3271663e6bfe0d69784164#diff-298f0bf26aaad42c9c163f69bba432beedac14f685c6c593957e46ae344ab480R81

I think the only thing missing would be player saved.

paulsnoops:
Thanks @JanEricNitschke :) I'm not too bothered about the save stat to be honest so unless you guys want it, this issue can be closed?

pnxenopoulos:
@paulsnoops Will close for now. You know where to find us if something doesn't work or if a feature crosses your mind :)

PR #203, opened 2022-11-24
Fixing issue with player statistics when a player changed their name mid game. Adding more metrics to player statistics

JanEricNitschke:
Fixes https://github.com/pnxenopoulos/awpy/issues/183

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,6 @@ dmypy.json
 
 # csgo demofiles
 .dem
+
+# vscode workspace file
+*code-workspace
--- a/awpy/analytics/stats.py
+++ b/awpy/analytics/stats.py
@@ -1,3 +1,24 @@
+"""Functions to calculate statistics for a player or team from a demofile.
+
+    Typical usage example:
+
+    from awpy.parser import DemoParser
+    from awpy.analytics.stats import player_stats
+
+    # Create the parser object.
+    parser = DemoParser(
+        demofile = "astralis-vs-liquid-m2-nuke.dem",
+        demo_id = "AST-TL-BLAST2019-nuke",
+        parse_frames=False,
+    )
+
+    # Parse the demofile, output results to a dictionary of dataframes.
+    data = parser.parse()
+    player_stats_json = player_stats(data["gameRounds"])
+    player_stats_json[76561197999004010]
+
+    https://github.com/pnxenopoulos/awpy/blob/main/examples/01_Basic_CSGO_Analysis.ipynb
+"""
 import pandas as pd
 
 # accuracy
@@ -13,8 +34,14 @@ def player_stats(game_rounds, return_type="json"):
         ct_side = r["ctSide"]
         t_side = r["tSide"]
         for p in ct_side["players"]:
-            if str(p["steamID"]) + " - " + p["playerName"] not in player_statistics:
-                player_statistics[str(p["steamID"]) + " - " + p["playerName"]] = {
+            if (
+                p["playerName"]
+                if p["steamID"] == 0
+                else str(p["steamID"]) not in player_statistics
+            ):
+                player_statistics[
+                    p["playerName"] if p["steamID"] == 0 else str(p["steamID"])
+                ] = {
                     "steamID": p["steamID"],
                     "playerName": p["playerName"],
                     "teamName": ct_side["teamName"],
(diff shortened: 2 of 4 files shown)


JanEricNitschke:
Addressed most points raised in https://github.com/pnxenopoulos/awpy/issues/182:

Added how many (0,1,2,3,4,5) kill rounds a player had as well as number of 1v(1,2,3,4,5) clutch attempts and successes.

Also added the number of traded deaths.

I also changed `all_true` to `any_true` for kast calculation because i think that more accurately describes what it is actually used for.

Lastly i moved all the stat finalizations into a singular loop instead of having one part stat to calculate/round.

paulsnoops:
@JanEricNitschke thank you for adding these features!

JanEricNitschke:
I know added the option to specify `CT` or `T` as `selected_side` to only consider things that happend while each player was playing on that side (like HLTV has that option)

Additionally i grouped some checks that were happening multiple times.

I am also now just taking the traded player from the trade kills `playerTradedSteamID` instead of going back to the last kill.
But i think the trade logic could be improved a bit in general. At the moment something is not considered a trade kill if there is an unrelated kill in between the death and the trade.

It is also not possible to have two players traded in one kill. Although that might actually be desired.

JanEricNitschke:
Changing the trade logic such that ONLY the trade time matters and not whether it was also the overall kill directly before the trade.

This increases (or leaves constant) the number of trade kills/traded deaths.

An example would be in the demo currently used for testing the stats module: https://www.hltv.org/matches/2337844/astralis-vs-liquid-blast-pro-series-global-final-2019

Here device dies in the 5th round (Score 2-2) to Twist in Lobby and is subsequently traded by Magisk within 2 seconds. (Default trade time is 5 seconds)

However as dupreeh gets a kill on NAF by Vent between twists kill and death this did not previously count as a traded death for device.

https://www.youtube.com/watch?v=yfwuNphy1wA

Also note that with the current logic a player can get a trade on themselves, netting them both a trade kill and a traded death, by killing their killer with a molotov or HE. This is the case both before and after the current changes.

With this change this also leads to a discrepancy of devices KAST for that match compared to what HLTV gives while it matches before, at least for him. However, i do not know if this is because of that or due to a difference in the trade time or something else. There is another example that adds a traded death and there the trade is a bit slower.

JanEricNitschke:
Fixed a bug and clarified the behaviour when trading enemies.

 A trade is always onto an enemy
If your teammate kills someone and then you kill them
-> that is not a trade kill for you
If you kill someone and then yourself
-> that is not a trade kill for you
  
Enemies CAN trade your own death
If you force an enemy to teamkill their mate after your death
 -> thats a traded death for you
If you force your killer to kill themselves (in their own molo/nade/fall)
-> that is a traded death for you

also added `playerTradedSide` to kill event

pnxenopoulos:
On line 844 in `parse_demo.go`, we could condense it further by just `return inTradeWindow(killA, killB, tickRate, tradeTime)` right?

The changes to the trade logic are well-welcomed :) -- thank you for spotting this.

pnxenopoulos:
As for how the trades match up with HLTV's, it's ok if there are discrepancies. I think as long as the logic is defensible and the results aren't too different, then it's fine. The main thing here is that trades are truly matched up to their associated kill events.

> Also note that with the current logic a player can get a trade on themselves, netting them both a trade kill and a traded death, by killing their killer with a molotov or HE. This is the case both before and after the current changes.

For me this is ok, but I can see why we may want to force a constraint that the player who killed cannot get the trade.

paulsnoops:
Just wanted to say I've been testing these changes and I have parsed 80+ demos and the stats are all looking good, trades are certainly better now, thank you

pnxenopoulos:
> Just wanted to say I've been testing these changes and I have parsed 80+ demos and the stats are all looking good, trades are certainly better now, thank you

Fantastic. Expect these to be rolled in today/tomorrow.

JanEricNitschke:
Note that in the merged verison in https://github.com/pnxenopoulos/awpy/pull/208 i have changed the naming for multi-kills and clutches from `3kills` to `kills3` (similar for clutches) to play nice with TypedDicts.

JanEricNitschke:
> On line 844 in `parse_demo.go`, we could condense it further by just `return inTradeWindow(killA, killB, tickRate, tradeTime)` right?
> 
> The changes to the trade logic are well-welcomed :) -- thank you for spotting this.

Changed that locally on the branch for the combined PR https://github.com/pnxenopoulos/awpy/pull/208.
Will wait with pushing in case there are more changes not rerun the building/testing pipeline unnecessarily.

PR #207, opened 2023-01-05
Added type hinting

JanEricNitschke:
I added type hinting to the current main branch.

I also expanded some function doc strings to include all the errors they raise where those were missing.

Additionally i expanded the github workflow to run linting with flake8 and type checking with mypy.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,18 @@ jobs:
           python -m pip install --upgrade pip
           pip install pytest
           pip install pytest-cov
+          pip install flake8
+          pip install mypy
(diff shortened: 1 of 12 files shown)


JanEricNitschke:
Expanded type hinting significantly. Mostly by creating TypedDicts for the json contents (Just copied from parse_demo.go structs).

For that i created an awpy.types module on the same level as awpy.utils.

Due to this i found a bug with the `frame_distance` function, fixed that and modified the test so it would catch that issue next time.

There is basically only one major todo and that would be to also build `TypedDict`s for the outputs of the `_parse_ABC` functions that produce dicts that contain some general info + pd.Dataframes

PR #215, opened 2023-01-22
Improvements to distance matrices

JanEricNitschke:
The main work in this PR is some added verbosity to the distance matrix calculations that now print which place/area they are currently calculating distances from.

Additionally i added a test that compares the contents of the precalculated place matrices with what the generate function produces. Differences highlight changes in the nav meshes or graph definition that should be investigated.

Lastly the median_dist section of the place distance matrices takes as long to compute as the full AREA_DIST_MATRIX while the other options are significantly faster. The median_dist is now set to 0.0 if no AREA_DIST_MATRIX is present to speed up the calculation.

### DIFF
--- a/awpy/analytics/nav.py
+++ b/awpy/analytics/nav.py
@@ -558,17 +558,17 @@ def generate_area_distance_matrix(map_name: str, *, save: bool = False) -> AreaM
             # Compute center of second area
             area2_x, area2_y, area2_z = _get_area_center(map_name, area2)
             # Calculate basic euclidean distance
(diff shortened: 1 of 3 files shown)


JanEricNitschke:
Adjusted `AREA_DIST_MATRIX` to have integer keys instead of str keys (str(integer)) to avoid the needless conversions.

This is backwards incompatible for people who are using them from a pervious version. (Which tbh is probably just me...)

pnxenopoulos:
We may want to rethink this as an artifact and as a new feature entirely. If you want me to put it on the [roadmap](https://github.com/users/pnxenopoulos/projects/5), please me know

PR #224, opened 2023-04-04
Updated go and demoinfocs-golang versions

JanEricNitschke:
The demoinfocs-golang upgrade required a minor change in the parser.

I also applied all recommendations from the linter like removing unsued functions/variables and compressing things like `my_bool == true`.


pnxenopoulos:
LGTM

PR #226, opened 2023-04-19
Adds MVP info

### LINKED ISSUE #221: Get round MVP
It seems that [demoinfocs-goland](https://github.com/markus-wa/demoinfocs-golang/blob/master/docs/game-events.md) has a _round_mvp_ event. Is possible to get this information from awpy?

agsalguero:
Fix #221

### DIFF
--- a/awpy/analytics/stats.py
+++ b/awpy/analytics/stats.py
@@ -291,6 +291,9 @@ def _is_clutch(
     total_players_victim_side = game_round[lower_side(victim_side) + "Side"]["players"]
     if total_players_victim_side is None:
         return False
+    # This gets messed up when a player disconnects (dies) in freeze time but
+    # reconnects in time to play the round.
+    # Then this triggers at only 3 "real" deaths.
     player_killed_victim_side = len(round_statistics["players_killed"][victim_side])
     return len(total_players_victim_side) - player_killed_victim_side == 1
 
@@ -299,7 +302,7 @@ def _find_clutcher(
     victim_side_players: list[Players],
     victim_side: Literal["CT", "T"],
     round_statistics: RoundStatistics,
-) -> str:
+) -> str | None:
     for player in victim_side_players:
         clutcher_key = (
             str(player["playerName"])
@@ -312,7 +315,7 @@ def _find_clutcher(
             and clutcher_key in round_statistics["active_players"]
         ):
             return clutcher_key
-    return ""
+    return None
 
 
 def _handle_clutching(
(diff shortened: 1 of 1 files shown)


awpy/types.py
```diff
@@ -482,6 +482,8 @@ class GameRound(TypedDict):
     weaponFires: Optional[list[WeaponFireAction]]
     flashes: Optional[list[FlashAction]]
     frames: Optional[list[GameFrame]]
+    mvp: Players
```
JanEricNitschke, line 485: I dont think the `gameRounds` dict contains a `Players` dict for the key `mvp` from your implementation.

I think it should have an `MVPSteamID: Optional[int]` and the same for MVPName as optional string.

JanEricNitschke (commented):
This already looks pretty great.

I just have a couple fairly small comments.

It would also be good if you could add test coverage to the parts you added on the python side.

Like checking the MVP scores in the stats module (download the reference demo and double check what you get with what is shown in the game and on hltv and then add that as a test).

You should also add a small test that the stats function works correctly if the MVPName and/or steamid are `None` or a bot is the mvp.

I think there should already be some examples that you can have a look at in that regard in the stats test file https://github.com/pnxenopoulos/awpy/blob/main/tests/test_stats.py but if you have any questions just ask.

awpy/parser/parse_demo.go
```diff
@@ -129,6 +129,9 @@ type GameRound struct {
 	WinningTeam          *string            `json:"winningTeam"`
 	LosingTeam           *string            `json:"losingTeam"`
 	Reason               string             `json:"roundEndReason"`
+	MVPName              string             `json:"MVPName"`
```
JanEricNitschke, line 132: I think according to the demoinfocs-golang documentation the player can be nil and so we should handle this like we do the other cases where that can happen.

https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v3@v3.3.0/pkg/demoinfocs/events#RoundMVPAnnouncement

https://github.com/pnxenopoulos/awpy/blob/main/awpy/parser/parse_demo.go#L205

ricardoruwer:
@agsalguero @JanEricNitschke let's go forward with this PR? 😄

JanEricNitschke:
@ricardoruwer There are still the open points i have mentioned + the PR has to be rebased to the current main. Sadly i dont have the time to do so at the moment.

But feel free to fork their repo and finish the PR from that.

agsalguero:
I'll have some time in a week or two. I'll take a look to it.

PR #228, opened 2023-04-24
Adding pre-commit hook with black and ruff and fixing ruff warnings.

JanEricNitschke:
The first commit is just the adding of the pre-commit config.

Afterwards i went through and ran each hook and all the files.

For ruff i am adding linting rule categories one by one and already address them.

The fixing of one of those has introduced type hint syntax that is only available from python 3.10 so i have raised the required python version. If not desired i can revert that but i think it should be fine to just support newer python versions.

I guess except they arent available on collab?

And for another fix, to the boolean trap, i have made a bunch of arguments keyword only. For example in the DemoParse constructor and in some plotting functions. In principle this is a backwards incompatible change and could break some user code. But i really hope that no one was passing arguments by position to:

```
def plot_positions(
    positions: Optional[list[tuple[float, float]]] = None,
    colors: Optional[list[str]] = None,
    markers: Optional[list[str]] = None,
    alphas: Optional[list[float]] = None,
    sizes: Optional[list[float]] = None,
    map_name: str = "de_ancient",
    map_type: str = "original",
    dark: bool = False,
    apply_transformation: bool = False,
)
```

or

```
  def __init__(
        self,
        demofile: str = "",
        outpath: Optional[str] = None,
        demo_id: Optional[str] = None,
        log: bool = False,
        parse_rate: int = 128,
        parse_frames: bool = True,
        parse_kill_frames: bool = False,
        trade_time: int = 5,
        dmg_rolled: bool = False,
        parse_chat: bool = False,
        buy_style: str = "hltv",
        json_indentation: bool = False,
    ):
```

This PR is not yet ready as there are some lint rules still missing and i have not updated the github actions.


JanEricNitschke:
I swapped the type checker from mypy to pyright because pyright supports `Literal math` while mypy does not and we are doing that a lot with the TypedDicts in `player_stats`.

JanEricNitschke:
Had to increase the required python version further to 3.11 because of another typing feature.

awpy/utils.py
```diff
@@ -1,16 +1,34 @@
...
 import re
 import subprocess
+from typing import Any
+
 import pandas as pd
+
 from awpy.types import Area
 
 
 class AutoVivification(dict):
```
pnxenopoulos, line 12: is defaultdict something we could use here?

pyproject.toml
```diff
@@ -3,14 +3,12 @@ requires = ["setuptools", "wheel"]
 
 [project]
 name = "awpy"
-authors = [
-    {name = "Peter Xenopoulos", email = "xenopoulos@nyu.edu"},
-]
+authors = [{ name = "Peter Xenopoulos", email = "xenopoulos@nyu.edu" }]
```
pnxenopoulos, line 6: add yourself as an author, if you are comfortable with being publicly listed

pnxenopoulos:
To deal with the Colab issue (I think they have 3.9), do you think we could provide a notebook that upgrades to 3.11?

pyproject.toml
JanEricNitschke: I will, thanks!

awpy/utils.py
JanEricNitschke: Yeah, i think think the Autovivification is not even used anywhere. I have a function that return a nested default dict like that already in the nav module.

https://github.com/pnxenopoulos/awpy/blob/main/awpy/analytics/nav.py#L370

I dont know if you want me to just delte the part here since it is currently only being used in nav or if i should replace this part with the definition from nav?

JanEricNitschke:
@pnxenopoulos It probably makes sense to include that in the colab linked in the README, right? Since that also shows how to install go and awpy.

I just tested it with this code and that worked:

```
#The code below installs 3.11 (assuming you now have 3.10) and restarts environment, so you can run your cells.

import sys #for version checker
import os #for restart routine

if '3.11' in sys.version:
  print('You already have 3.11, nothing to install')
elif '3.10' in sys.version:
  print("Python version is: ", sys.version)

  print("Printing content of /usr/local/lib/python* to see available versions")
  !ls /usr/local/lib/python*

  #install python 3.11 and dev utils
  #you may not need all the dev libraries, but I haven't tested which aren't necessary.
  !sudo apt-get update -y > /dev/null
  !sudo apt-get install python3.11 python3.11-dev python3.11-distutils libpython3.11-dev > /dev/null
  !sudo apt-get install python3.11-venv binfmt-support  > /dev/null #recommended in install logs of the command above

  #change alternatives
  !sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 1 > /dev/null
  !sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.11 2 > /dev/null

  # install pip
  !curl -sS https://bootstrap.pypa.io/get-pip.py | python3.11  > /dev/null

  #install colab's dependencies
  !python3 -m pip install ipython==7.9.0 traitlets==5.7.1 jupyter psutil matplotlib setuptools ipython_genutils ipykernel jupyter_console notebook prompt_toolkit httplib2 astor  > /dev/null

  #minor cleanup
  !sudo apt autoremove > /dev/null

  #link to the old google package
  !ln -s /usr/local/lib/python3.10/dist-packages/google /usr/local/lib/python3.11/dist-packages/google > /dev/null

  #!jupyter serverextension list  > /dev/null
  #!jupyter notebook --config=/usr/local/etc/jupyter/jupyter_notebook_config.json  > /dev/null
  
  #this is just to verify if 3.11 folder was indeed created
  print("Printing content of /usr/local/lib/python3.11/")
  !ls /usr/local/lib/python3.11/

  #restart environment so you don't have to do it manually
  os.kill(os.getpid(), 9)
else:
  print("Your out of the box Pythong is not 3.10, so probably the script will not work, so pls feel free to edit the script to ignore then check and re-run: ", sys.version)
```

JanEricNitschke:
I think it makes sense to merge this one first and then let me rebase all the others and potentially fix their linting warnings (and conflicts)

JanEricNitschke:
I corrected the `package_data` in setup.py to include the distance matrices, i moved the pytest and coverage settings from setup.cfg (and deleted it) to pyproject.toml and i tried to also specify the package_data in pyproject.toml. But im not sure if i did it correctly and how one could easily test that.

Issue #193, opened 2022-10-03
Add frame id


pnxenopoulos:
Add frame id to the frame top-level

PR #229, opened 2023-04-25
Fixing issue with steamids being changed due to float conversion.

JanEricNitschke:
Addresses https://github.com/pnxenopoulos/awpy/issues/213.

I also added a test for exactly this and cleaned up the `_parse_kills`... functions.

PR #230, opened 2023-04-26
Adding per round frameID entry to GameFrames.

JanEricNitschke:
Adding `frameID` key to each GameFrame that contains the index of the frame in its round. Addresses: https://github.com/pnxenopoulos/awpy/issues/193

This means that 

`[frame["frameID"] for frame in game_round["frames"]] == list(range(len(game_round["frames"])))`

or 

```
for index, frame in enumerate(game_round["frames"]):
    index == frame["frameId"]
```

I dont know if a "global" frameID would also be interesting. That would be a tiny bit harder to implement.

### DIFF
--- a/awpy/parser/demoparser.py
+++ b/awpy/parser/demoparser.py
@@ -898,7 +898,7 @@ class DemoParser:
 
         Raises:
             AttributeError: Raises an AttributeError if the .json attribute
-                has a "gameRounds" key.
+                has no "gameRounds" key.
         """
         if self.json and self.json["gameRounds"]:
             for i, _r in enumerate(self.json["gameRounds"]):
--- a/awpy/parser/demoparser.py
+++ b/awpy/parser/demoparser.py
@@ -869,6 +869,7 @@ class DemoParser:
             if remove_bad_scoring:
                 self.remove_bad_scoring()
             self.renumber_rounds()
+            self.renumber_frames()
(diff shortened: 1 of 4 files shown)


JanEricNitschke:
Test failures see: https://github.com/pnxenopoulos/awpy/pull/229#issuecomment-1522360733

JanEricNitschke:
I am now trying to add a global frame index like this:

```
func appendFrameToRound(currentRound *GameRound, currentFrame *GameFrame, globalFrameIndex *int64) {
	currentFrame.FrameID = int64(len(currentRound.Frames))
	currentFrame.GlobalFrameID = int64(*globalFrameIndex)
	*globalFrameIndex += 1
	currentRound.Frames = append(currentRound.Frames, *currentFrame)
}
```

But my test of:

```
       for index, frame in enumerate(
            frame
            for game_round in self.index_parser.json["gameRounds"]
            for frame in game_round["frames"]
        ):
            assert index == frame["globalFrameID"]
```

keeps failing because my indices always start at some random non-zero number before counting up as expected.

I guess it is somewhere adding frames to rounds where either the frame or the round is later deleted or the round is not added to the game?

JanEricNitschke:
Managed to solve the problem. Now there is a functioning global frame index.

However at the moment it still gets messed up during cleaning. Will add a function fixing that to the python side cleaning logic in the same way that the round renumbering is done after cleaning.

PR #231, opened 2023-04-27
Changing DemoParser signature to use **kwargs.

JanEricNitschke:
Changed the DemoParser `__init__` function to use `**parser_args` for all but the first 4 arguments.

```
        demofile: str = "",
        outpath: Optional[str] = None,
        demo_id: Optional[str] = None,
        log: bool = False,
```

Properly type hinting that change either requires Python3.11 or to add typing_extensions to the dependencies. I have currently included the first option but do not have a string opinion on this.


pnxenopoulos:
I think with Python 3.11 it is fine, that seems to be where things are headed. As long as we can sort out Colab, I think it will be fine.

JanEricNitschke:
Now all the commits are there but the CI didnt run.

PR #232, opened 2023-04-28
Refactoring go code

JanEricNitschke:
Mainly decluttering the main function by moving each even handler registration into a separate function.

There should currently only be one functional change. Namely that `RoundEndOfficial` events do not overwrite `EndTick` if it was properly set by `RoundEnd`

Also planning to add some tests where easily possible.


JanEricNitschke:
This is now also ready for a review.

There is still some stuff do to, but that is outside the scope of this PR.

For example there are still some extremely complex functions and there is also a lot of duplication with BombEvent handling.

pnxenopoulos (approved):

PR #233, opened 2023-04-30
Adding more CI configurations

JanEricNitschke:
Added configurations to run CI also on windows and mac and for python and go version from the required up to the most recent.


JanEricNitschke:
Once the other PRs are merged i will rebase this onto them and then want to address all of the (remaining) pylint warnings.

I think there are a bunch of refactoring opportunities in my nav code that can make it more readable.
I also think that the cleaning can be improved so that we dont have to iterate over all rounds for each cleaning step but can just iterator once.

JanEricNitschke:
Fixing some pylint warnings. Did some refactoring to do that which changed the signature of `plot_position` significantly in a backwards incompatible way.

https://github.com/pnxenopoulos/awpy/pull/233/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316R130-R137

JanEricNitschke:
Will do some more in that direction with DemoParser arguments and also some more refactoring in nav module until all pylint warnings are gone. Then we probably need a new minor release because of the new requirements and backwards incompatible changes.

JanEricNitschke:
Added badges for the formatter and linters we are using.

I wanted to add a coverage badge according to [this])(https://nedbatchelder.com/blog/202209/making_a_coverage_badge.html) but i dont have the necessary permissions.

And sadly pyright does not have a badge to show for the type checking.

PR #234, opened 2023-05-01
DRAFT: Trying stuff out for shared library for go code.

JanEricNitschke:
I managed to get [gopy](https://github.com/go-python/gopy) running locally for me on windows so that i can create a library and it works. But there is still a bunch of stuff i am not sure about.


JanEricNitschke:
If i just push the compiled library from my windows machine the windows workflows work. Makes sense.

But i dont know how to solve this when we want to push this via pip. Because to compile it from source go needs to be installed which kinda defeats the whole purpose. And i dont think there is a single way we can precompile this that works on all machines.

Maybe we have to include library for both windows and linux (and mac?)?

JanEricNitschke:
Adding more resources for when i can get back to this: https://last9.io/blog/using-golang-package-in-python-using-gopy/

Issue #236, opened 2023-05-19
awpy.visualization.plot.plot_round: TypeError: The keyword `fps` is no longer supported. 


gerelef:
**Describe the bug**
Ran something similar to the example provided [here](https://github.com/pnxenopoulos/awpy/blob/main/examples/02_Basic_CSGO_Visualization.ipynb).
Expected a gif output of player movements, however it threw an exception and exited abnormally. 
From the detailed description provided by `pillow`, I gather that the formerly `fps` value has been removed and been replaced by `duration`.

**Snippet:**
```python3
import awpy
from awpy.visualization.plot import plot_round

data = awpy.DemoParser(
    demofile="vitality-vs-into-the-breach-m1-vertigo.dem"
).parse()

plot_round("best_round_ever.gif", data["gameRounds"][7]["frames"], map_name=data["mapName"], map_type="simpleradar", dark=False)
```

**Additional context**
Linux, Fedora 37.
Latest version of awpy from pip as of writing.
go version go1.19.9 linux/amd64

Match demo is [here](https://www.hltv.org/matches/2364289/vitality-vs-into-the-breach-blasttv-paris-major-2023).

**Stacktrace:**
```python3
80it [00:50,  1.58it/s]
Traceback (most recent call last):
  File "/home/user/random/awpy-test/main.py", line 9, in <module>
    plot_round("best_round_ever.gif", data["gameRounds"][7]["frames"], map_name=data["mapName"], map_type="simpleradar", dark=False)
  File "/home/user/random/awpy-test/venv/lib64/python3.11/site-packages/awpy/visualization/plot.py", line 257, in plot_round
    imageio.mimsave(filename, images, fps=fps)
  File "/home/user/random/awpy-test/venv/lib64/python3.11/site-packages/imageio/v2.py", line 484, in mimwrite
    return file.write(ims, is_batch=True, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/random/awpy-test/venv/lib64/python3.11/site-packages/imageio/plugins/pillow.py", line 354, in write
    raise TypeError(
TypeError: The keyword `fps` is no longer supported. Use `duration`(in ms) instead, e.g. `fps=50` == `duration=20` (1000 * 1/50).```

JanEricNitschke:
[...]

PR #237, opened 2023-05-19
Fixing requirements to tighter bounds

JanEricNitschke:
To avoid regressions from updated dependencies like in https://github.com/pnxenopoulos/awpy/issues/236


JanEricNitschke:
Check go failure is from here: https://github.com/golangci/golangci-lint/issues/386
and unrelated to any changes from us.

PR #252, opened 2023-06-10
fixed test locations

pnxenopoulos:
Changed the test demo locations to mediafire to be easier on my GCP budget

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,11 +1,14 @@
-import pytest
+"""Global test configuration."""
 import json
-import requests
 import os
+
+import pytest
+import requests
 
 
 @pytest.fixture(scope="session")
-def setup():
+def setup() -> None:  # noqa: PT004
+    """Sets up testing environment by downloading demofiles."""
     with open("tests/test_data.json", encoding="utf-8") as f:
         demo_data = json.load(f)
     for file in demo_data:
@@ -13,7 +16,8 @@ def setup():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def teardown():
+def teardown() -> None:  # noqa: PT004
+    """Cleans testing environment by deleting all .dem and .json files."""
     files_in_directory = os.listdir()
     if filtered_files := [
         file for file in files_in_directory if file.endswith((".dem", ".json"))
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+    with open("tests/test_data.json", encoding="utf-8") as f:
+        demo_data = json.load(f)
+    for file in demo_data:
+        _get_demofile(demo_link=demo_data[file]["url"], demo_name=file)
+
+
+@pytest.fixture(scope="module", autouse=True)
+def teardown() -> None:  # noqa: PT004
+    """Cleans testing environment by deleting all .dem and .json files."""
+    files_in_directory = os.listdir()
```
JanEricNitschke, line 21: I think you need to add a `yield` here so that the tests are run before this part. Otherwise i think the function will run instantly and just delete the downloaded files?

JanEricNitschke: Changing the scope of the two fixtures back to `session` and adding a `yield` to the start of the teardown gets it to work for me locally.

But even with this it will still get downloaded for every build in the matrix and every new PR/commit.

JanEricNitschke:
I think if we add this to the testing workflow we could cut down on the downloads (per [here):](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action)

```
      - name: Cache test demos
        id: cache-demos
        uses: actions/cache@v3
        env:
          cache-name: cache-demos
        with:
          # demos are downloaded to
          # and looked for in `~/.npm`
          path: ${{ github.workspace }}/tests/*.dem
          key: cache-test-demos-${{ hashFiles('**/test_data.json') }}
          restore-keys: cache-test-demos-
```

Then we should just check [here](https://github.com/pnxenopoulos/awpy/blob/8baba835be453ac0ceeb6273fc3cf66d66c029ad/tests/conftest.py#L36C37-L36C37) if the file already exists and only download it if not.

And if we are already changing the file anyway we can change [this](https://github.com/pnxenopoulos/awpy/blob/8baba835be453ac0ceeb6273fc3cf66d66c029ad/.github/workflows/build.yml#L21) to say "Checkout awpy library"

tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+    Args:
+        demo_link (str): Link to demo.
+        demo_name (str): `<file>.dem` styled filename.
+    """
+    if os.path.exists(f"tests/{demo_name}.dem"):
+        pass
+    else:
+        r = requests.get(demo_link, timeout=100)
+        with open(f"tests/{demo_name}.dem", "wb") as demo_file:
+            demo_file.write(r.content)
```
JanEricNitschke, line 42:
```suggestion
    if not os.path.exists(demo_path := f"tests/{demo_name}.dem"):
        r = requests.get(demo_link, timeout=100)
        with open(demo_path, "wb") as demo_file:
            demo_file.write(r.content)
```

I think i would make this one change to avoid typing the demo_path twice and reduce the chance of typos.

In general this looks really awesome now and should make the tests way cleaner and also faster.

PR #253, opened 2023-06-20
Adds MVP info 

agsalguero:


awpy/parser/parse_demo.go
```diff
@@ -136,6 +136,9 @@ type GameRound struct {
 	WinningTeam          *string            `json:"winningTeam"`
 	LosingTeam           *string            `json:"losingTeam"`
 	Reason               string             `json:"roundEndReason"`
+	MVPName              string             `json:"MVPName"`
+	MVPSteamID           int64              `json:"MVPSteamID"`
+	MVPReason            string             `json:"MVPReason"`
```
JanEricNitschke, line 141: I think according to the demoinfocs-golang documentation the player can be nil and so we should handle this like we do the other cases where that can happen.

https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v3@v3.3.0/pkg/demoinfocs/events#RoundMVPAnnouncement

https://github.com/pnxenopoulos/awpy/blob/main/awpy/parser/parse_demo.go#L205

awpy/parser/parse_demo.go
```diff
@@ -1461,6 +1477,14 @@ func registerRoundEndHandler(demoParser *dem.Parser, currentGame *Game, currentR
 	})
 }
 
+func registerRoundMVPHandler(demoParser *dem.Parser, currentGame *Game, currentRound *GameRound) {
+	(*demoParser).RegisterEventHandler(func(e events.RoundMVPAnnouncement) {
+	    currentRound.MVPName = e.Player.Name
+	    currentRound.MVPSteamID = int64(e.Player.SteamID64)
+	    currentRound.MVPReason = convertRoundMVPReason(e.Reason)
+	})
```
JanEricNitschke, line 1485: I think to be clean we need a nil check here for `e.Player`?

awpy/parser/parse_demo.go
```diff
@@ -1461,6 +1477,14 @@ func registerRoundEndHandler(demoParser *dem.Parser, currentGame *Game, currentR
 	})
 }
 
+func registerRoundMVPHandler(demoParser *dem.Parser, currentGame *Game, currentRound *GameRound) {
```
JanEricNitschke, line 1480: Linter says that `currentGame` is unused here. Can you have a look.

JanEricNitschke (commented):
Hi, thanks for your contribution.

Looks very good to me already. I just added some small comments.

I also see that the python part is failing the linting.

We have established a fairly thorough formatting/linting/testing process using black, ruff, pyright, pylint and pytest.

https://github.com/pnxenopoulos/awpy/blob/main/.github/workflows/build.yml

Some of these checks you can very easily do locally via pre-commit:
https://github.com/pnxenopoulos/awpy/blob/main/.pre-commit-config.yaml
https://pre-commit.com/

For the others you would have to manually install pyright and pylint or use the CI to check if the pass.

It would be great if you could adjust your code so that it passes all of the linting steps.

awpy/parser/parse_demo.go
agsalguero: Done

awpy/parser/parse_demo.go
pnxenopoulos: Indeed, the comment by @JanEricNitschke is correct. We can achieve this simply by using the `*` to denote a pointer.

awpy/types.py
```diff
@@ -548,6 +548,9 @@ class GameRound(TypedDict):
     winningTeam: str | None
     losingTeam: str | None
     roundEndReason: str
+    MVPName: str | None
```
pnxenopoulos, line 551: for naming convention purposes, we can make these `mvpName`, `mvpSteamId`, and so on

pnxenopoulos (changes requested):
Really great work on this one! I appreciate the edits on the notebooks as well. I only have a few small comments that need to be addressed, then we can merge this one into `main`:

- in the `GameRound` struct, change `MVPName`, `MVPSteamID`, and `MVPReason` to be pointers by using `*`
- in the awpy types file, use lowercase casing (e.g. `mvpName`) for the new columns

pnxenopoulos (approved):

awpy/analytics/stats.py
```diff
@@ -620,6 +621,21 @@ def _handle_multi_kills(
             _increment_statistic(player_statistics, player, n_kills)
 
 
+def _handle_mvps(
+    game_round: GameRound,
+    player_statistics: dict[str, PlayerStatistics],
+    round_statistics: RoundStatistics,
+) -> None:
+    if "mvpName" in game_round:
```
JanEricNitschke, line 629: That check should be unnecessary since we per type definition a "GameRound" should always have that key.

But it probably does make sense to leave this in to allow the stat extraction of old json files that were produced with a previous version.

awpy/analytics/stats.py
```diff
@@ -620,6 +621,21 @@ def _handle_multi_kills(
...
+        player_key = (
```
JanEricNitschke, line 630: This could probably make use of `_get_actor_key`

awpy/parser/parse_demo.go
```diff
@@ -556,6 +559,19 @@ func convertRoundEndReason(r events.RoundEndReason) string {
 	}
 }
 
+func convertRoundMVPReason(r events.RoundMVPReason) string {
```
JanEricNitschke, line 562: This is probably one of the few functions that we can actually easily test on the go side. I feel like we should just do it.

awpy/types.py
```diff
@@ -569,6 +569,9 @@ class GameRound(TypedDict):
     winningTeam: str | None
     losingTeam: str | None
     roundEndReason: str
+    mvpName: str | None
```
JanEricNitschke, line 572: If we do NOT make them pointers on the go side then we can remove the `| None` here.

I still think these should be null of no player exists for the MVP announcement but if we do not want to go that route than we should still be consistent and remove it here.

JanEricNitschke (commented):
I added some comments.

The main point is whether we should have the mvp name and steamid nullable or not. I feel that we do.

awpy/parser/parse_demo.go
```diff
@@ -568,7 +568,7 @@ func convertRoundMVPReason(r events.RoundMVPReason) string {
 	case events.MVPReasonBombPlanted:
 		return "MVPReasonBombPlanted"
 	default:
-		return unknown
+		return "Unknown"
```
JanEricNitschke, line 571: I think that should stay as `unkown` which is defined here: https://github.com/pnxenopoulos/awpy/blob/main/awpy/parser/parse_demo.go#L19

As one of the linters was complaining about it.

JanEricNitschke:
What is with the `golangci-lint` binary file?

JanEricNitschke:
Tests fail because the manual game_rounds do not include that key:
https://github.com/pnxenopoulos/awpy/blob/3945c2ee521ef055c921b71883909a71821ec85b/tests/test_stats.py#L182

It might actually make sense to keep that check to not break on old parsed jsons.

Issue #254, opened 2023-07-04
nil pointer dereference when parsing POV demos recorded from coach perspective


emteka:
**Describe the bug**
I can't get full output out of demo parser. It get's stuck on some nil value and terminates the process, leaving file unfinished.

Here's example console error:
```
C:\Users\mtk\PycharmProjects\pythonProject\venv\Scripts\python.exe C:\Users\mtk\PycharmProjects\pythonProject\main.py 
panic: runtime error: invalid memory address or nil pointer dereference
stacktrace:
goroutine 53 [running]:
runtime/debug.Stack()
	C:/Program Files/Go/src/runtime/debug/stack.go:24 +0x65
github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs.NewParserWithConfig.func1({0x842060, 0xc0009294a0})
	C:/Users/mtk/go/pkg/mod/github.com/markus-wa/demoinfocs-golang/v3@v3.3.0/pkg/demoinfocs/parser.go:362 +0x30
github.com/markus-wa/godispatch.(*Dispatcher).dispatchWithRecover.func1()
	C:/Users/mtk/go/pkg/mod/github.com/markus-wa/godispatch@v1.4.1/dispatch.go:197 +0x42
panic({0x842060, 0xc0009294a0})
	C:/Program Files/Go/src/runtime/panic.go:884 +0x213
github.com/markus-wa/godispatch.callConsumerCode.func1()
	C:/Users/mtk/go/pkg/mod/github.com/markus-wa/godispatch@v1.4.1/dispatch.go:122 +0x5f
panic({0x842060, 0xc000929480})
	C:/Program Files/Go/src/runtime/panic.go:884 +0x213
github.com/markus-wa/godispatch.callConsumerCode.func1()
	C:/Users/mtk/go/pkg/mod/github.com/markus-wa/godispatch@v1.4.1/dispatch.go:122 +0x5f
panic({0x81f100, 0xb743d0})
	C:/Program Files/Go/src/runtime/panic.go:884 +0x213
github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs/common.(*Equipment).UniqueID(...)
	C:/Users/mtk/go/pkg/mod/github.com/markus-wa/demoinfocs-golang/v3@v3.3.0/pkg/demoinfocs/common/equipment.go:315
[...]

PR #255, opened 2023-07-04
fix some nils for POV demos

emteka:
fix some nils for POV demos

### DIFF
--- a/awpy/parser/parse_demo.go
+++ b/awpy/parser/parse_demo.go
@@ -1135,8 +1135,12 @@ func registerRoundStartHandler(demoParser *dem.Parser, currentGame *Game, curren
 		currentRound.IsWarmup = gs.IsWarmupPeriod()
 		currentRound.RoundNum = int64(len(currentGame.Rounds) + 1)
 		currentRound.StartTick = int64(gs.IngameTick())
-		currentRound.TScore = int64(gs.TeamTerrorists().Score())
-		currentRound.CTScore = int64(gs.TeamCounterTerrorists().Score())
+		if (gs.TeamTerrorists() != nil) {
+		    currentRound.TScore = int64(gs.TeamTerrorists().Score())
+		}
+		if (gs.TeamCounterTerrorists() != nil) {
+		    currentRound.CTScore = int64(gs.TeamCounterTerrorists().Score())
+		}
 		if (gs.TeamTerrorists() != nil) && (gs.TeamCounterTerrorists() != nil) {
 			tTeam := gs.TeamTerrorists().ClanName()
 			ctTeam := gs.TeamCounterTerrorists().ClanName()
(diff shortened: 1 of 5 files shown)


JanEricNitschke:
I would combine these lines: https://github.com/emteka/awpy-1/blob/fix-nils-in-demo-parser/awpy/parser/parse_demo.go#L1138-L1149

And probably move them into their own function as there are identical checks here: https://github.com/emteka/awpy-1/blob/fix-nils-in-demo-parser/awpy/parser/parse_demo.go#L1200-L1207
and here: https://github.com/emteka/awpy-1/blob/fix-nils-in-demo-parser/awpy/parser/parse_demo.go#L1264-L1271

Also this part seems completely unnecessary

Also protect these within nil checks.

JanEricNitschke:
In relation to https://github.com/pnxenopoulos/awpy/issues/254

JanEricNitschke:
With the changes now a score of 0 is used for a team if it could not be extracted.
This means that `remove_bad_scoring` will effectively delete all of the rounds during cleaning. By default this is one.

This makes it so that for POV demos the cleaning will just remove all the rounds without it being directly clear what the cause is.

I have two ideas for how to address this. 

The first would be to add a logging statement when `remove_bad_scoring` removes ALL rounds (maybe this makes sense to to for all individual cleaning functions?).

The second would be to change the default value of that toggle depending on the `clientName` in the json. For normal GOTV demos it is "GOTV Demo" where as for POV demos it is usually the person recording them.

Right now i am leaning towards the first and i think i could do it by writing a decorator that we can apply to the individual cleaning functions.

pnxenopoulos:
> With the changes now a score of 0 is used for a team if it could not be extracted. This means that `remove_bad_scoring` will effectively delete all of the rounds during cleaning. By default this is on. https://github.com/pnxenopoulos/awpy/blob/main/awpy/parser/demoparser.py#L876
> 
> This makes it so that for POV demos the cleaning will just remove all the rounds without it being directly clear what the cause is.
> 
> I have two ideas for how to address this.
> 
> The first would be to add a logging statement when `remove_bad_scoring` removes ALL rounds (maybe this makes sense to to for all individual cleaning functions?).
> 
> The second would be to change the default value of that toggle depending on the `clientName` in the json. For normal GOTV demos it is "GOTV Demo" where as for POV demos it is usually the person recording them.
> 
> Right now i am leaning towards the first and i think i could do it by writing a decorator that we can apply to the individual cleaning functions.
> 
> Edit: Nevermind, it doesnt actually delete all the rounds. But might still be a sensible thing to do with the decorator.

What's your concern with the second option?

JanEricNitschke:
> What's your concern with the second option?

I am not sure if it is 100% reliable and it is a bit opaque.

But from looking at the actual examples it doesnt even seem to be the case. It seems that the opposite team is nil only for some events (did not test which) and the later events correctly overwrite the score.

So after the fix the scores seemed somewhat reasonable, although they were still weird with games going past 15 (not OT) or shorter. I did not check if that is an error with the parsing or due to custom practice game settings.

JanEricNitschke:
> Please keep in mind - in case of POV demo from PCW going over 15 without ending the match is completely normal - teams agree on playing 30 round match no matter the result, so in some cases we might have matches like 25:5

I thought something like that. I will double check one of the demos you have linked in your issue to see if the current implementation handles it correctly.

Then i would probably also write a test for that. @pnxenopoulos you would have to upload it to figshare and update the `test_data.json`

PR #257, opened 2023-07-08
Updated linters.

JanEricNitschke:
Updated linters in CI and pre-commit.

Added perflint rules to ruff configuration and fixed one raised warning.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,13 +65,11 @@ jobs:
       - name: Lint with ruff
         uses: chartboost/ruff-action@v1
         with:
-          version: 0.0.267
-
+          version: 0.0.277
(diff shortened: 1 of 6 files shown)


pnxenopoulos:
Is there any argument for just letting these default to the current release versions?

JanEricNitschke:
We could default to most recent for the CI but one has to manually update the local versions from pre-commit. This can be annoying when the linters make breaking changes that require larger work and this only gets noticed in the CI of an otherwise small PR.

We had that twice already with pyright and golangcilint updates. It took quite a bit of time to understand that the CI issue wasnt from the code changes but instead because this CI run used newer versions of the linters. It can also cause situations where the CI passes in the last PR commit but then fails after the merge of the main branch.

PR #258, opened 2023-07-08
Added pydantic check to setting of parser json. Updated requirements.

JanEricNitschke:
Added a check with pydantic that the loaded json conforms to the expected structure from the awpy.types module.

If the check fails a warning is emitted but the program continues.

Due to the check if fixed some typos in the definitions of the typed dicts.

At the moment usage of pydantic requires typeddict to be imported from typing_extensions.
So i have added that as a requirement and disabled the ruff warning that wants me to change it back.

Also updated/fixed some requirement specifications.

### DIFF
--- a/awpy/parser/demoparser.py
+++ b/awpy/parser/demoparser.py
@@ -214,21 +214,21 @@ class DemoParser:
         return self._json
 
     @json.setter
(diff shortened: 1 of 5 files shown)


JanEricNitschke:
Want to set `strict=True`

PR #263, opened 2023-07-24
DRAFT: HLTV Win Probability Matrix 

hojlund123:
Finally got the datacleaning done. Alot of it is still on testing basis but need some suggestions on additional helpers and how we should structure it.


JanEricNitschke:
I'll grab the branch and check it out later.

But for general thought. We should probably have a an existing version of the matrix stored as a json file.

I am not sure how to handle updating it yet. 
Options would be:
- on install
- on call of a separate function
- on grabbing a value from the matrix (always)
- on grabbing a value from the matrix (if the matrix is old)
- on grabbing a value from the matrix (when a boolean arg is specified)

if trying to update then notify the user of success or failure.
if we update during grabbing and fail just grab the old value

With this change request would be a new runtime requirement now i think.
Should we make that optional on installation? (Does anyone know how that works?)

- When downloading i think we should handle the cleaning in code and only write the final matrix as json.

We definitely have to pass the CI.

- Pass formatting from black
- Pass lints from ruff
- Pass type checking from pyright:
- -> Define a pydantic model/typeddict (and check with pydantic) that we expect the matrix structure to follow
- Pass pylint

Names shoud probably be more descriptive

JanEricNitschke:
For the CI we are using pre-commit which should run black and ruff for you and then you would have to manually install and run pyright and pylint. 

Ill try to write a contributing file soon.

awpy/analytics/wpa.py
```diff
@@ -30,9 +31,57 @@ def round_win_probability(ct_score: int, t_score: int, map_name: str) -> dict:
...
+        map_name (str): Map name the demo is from
 
     Returns:
-        A dictionary containing the CT game win, T game win and Draw probabilities
+        A dictionary containing the game win probability
     """
-    raise NotImplementedError
+    # Load the data from the json file
+    wpa_data = json.loads("../data/wpa/wpa.json")
+
```
JanEricNitschke, line 41: I think the `wpa_data should be loaded in the data `__init__` file and just imported in wpa.py.

awpy/analytics/wpa.py
```diff
@@ -30,9 +31,57 @@ def round_win_probability(ct_score: int, t_score: int, map_name: str) -> dict:
...
+    ):
+        return {"error": "Invalid input"}
+
+    # Get the win probability
+    win_prob = wpa_data[str(map_id)][team][str(ct_score)][str(t_score)]
+
+    return {
+        "win_probability": win_prob,
+    }
+
```
JanEricNitschke, line 63: - I do not quite get the thing about determining the team from the score?
- Shouldnt the CT win probability just be data[CT][ct_score][t_score] and the T probability be data[T][ct_score][t_score]
- I think we can just return a dict with the ct, t and draw probability.

We really do not need to return a dict with a singular item.

awpy/analytics/wpa.py
```diff
@@ -30,9 +31,57 @@ def round_win_probability(ct_score: int, t_score: int, map_name: str) -> dict:
...
+    if (
+        str(map_id) not in wpa_data
+        or team not in wpa_data[str(map_id)]
+        or str(ct_score) not in wpa_data[str(map_id)][team]
+        or str(t_score) not in wpa_data[str(map_id)][team][str(ct_score)]
+    ):
+        return {"error": "Invalid input"}
+
+    # Get the win probability
+    win_prob = wpa_data[str(map_id)][team][str(ct_score)][str(t_score)]
```
JanEricNitschke, line 58: I think this should just raise an error.

awpy/analytics/wpa.py
```diff
@@ -30,9 +31,57 @@ def round_win_probability(ct_score: int, t_score: int, map_name: str) -> dict:
...
+        "de_inferno": 33,
+        "de_nuke": 34,
+        "de_train": 35,
+        "de_cbble": 39,
+        "de_overpass": 40,
+        "de_vertigo": 46,
+        "de_ancient": 47,
+        "de_anubis": 48,
+    }
+
```
JanEricNitschke, line 84: Are there more maps on hltv?

awpy/data/wpa/jsparser.py
```diff
@@ -0,0 +1,57 @@
+"""Module to fetch and clean WPA data from the HLTV website."""
+
+import ast
+import json
+
+import requests
+
+URL = "https://www.hltv.org/scripts/hltv.js"
```
JanEricNitschke, line 8:
```suggestion
HLTV_URL = "https://www.hltv.org/scripts/hltv.js"
```

awpy/data/wpa/jsparser.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def fetch_matrix_data() -> dict:
+    """Function to get the latest matrix data from the HLTV.
```
JanEricNitschke, line 12:
```suggestion
    """Function to get the latest matrix data from HLTV.
```

awpy/data/wpa/jsparser.py
```diff
@@ -0,0 +1,57 @@
...
+        match err:
+            case requests.exceptions.HTTPError as http_err:
+                print(f"HTTP error occurred: {http_err}")
+            case requests.exceptions.ConnectionError as conn_err:
+                print(f"Connection error occurred: {conn_err}")
+            case requests.exceptions.Timeout as timeout_err:
+                print(f"Timeout error occurred: {timeout_err}")
+            case requests.exceptions.RequestException as req_err:
+                print(f"Other request error occurred: {req_err}")
+        return {}  # return an empty dict in case of error
```
JanEricNitschke, line 54: I think we should use proper logging here.

JanEricNitschke, line 54: I think this should either (reraise the error/ just not catch them), the preexisting data or the function should just not return anything at all.

JanEricNitschke (commented):
Nice job already. I left a couple of comments. 

It also seems that the formatting is still not 100% right and black is complaining.

Another thing we have to figure out is when to call the fetch function (and how to get this all legally correcet)

JanEricNitschke:
We should also definitely make this all type save with type dicts and/or pydantic.

Additionally i am currently getting a `JSONDecodeError` when trying to call `round_win_probability`.

So we should also definitely add tests.

awpy/analytics/wpa.py
JanEricNitschke: It seems there is also dust2 with 31. And ancient is actually missing in the .js file.

awpy/analytics/wpa.py
JanEricNitschke: And the structure should be validated on loading (so at the top or ideally in data init) and updating (so when requesting and cleaning it from htlv)

awpy/analytics/wpa.py
hojlund123: Good idea!

awpy/analytics/wpa.py
hojlund123: I was trying to do some further calcs but figured we don't need all that. Will make it much simpler like:

`# Get the win probability from the loaded data
    ct_win_probability = wpa_data[str(map_id)]["CT"][str(ct_score)][str(t_score)]
    t_win_probability = wpa_data[str(map_id)]["T"][str(ct_score)][str(t_score)]

    return {"CT": ct_win_probability, "T": t_win_probability}`

PR #265, opened 2023-07-24
Update ci

JanEricNitschke:
Testing out new pre-commit hook and trying to add go fmt check to go linters.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Typecheck with pyright
         uses: jakebailey/pyright-action@v1
         with:
-          version: 1.1.318
+          version: 1.1.319
 
       - name: Thorough check with pylint
         run: pylint awpy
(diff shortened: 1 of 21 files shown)


JanEricNitschke:
I think i finally managed to also add the golang-ci linter into the pre-commit hooks.

JanEricNitschke:
Also added pyright and pylint to be run as a pre-commit hook. They still have to be installed separately (because they need access to the normal environment) but now they do not have to be run manually anymore.

pnxenopoulos (approved):
Looks good to me. Was there a reason we had pylint running with `fail-under` at 9.99 before?

JanEricNitschke:
That was from when i added it. I think i added it with 9.95 and then gradually increased it as i was fixing the original set of warnings. And i think when we merged the first set there was still a PR open that would fix that last bit already and so it wasnt in the original PR itself.

And then i just forgot to increase it to 10.0 afterwards.

PR #268, opened 2023-07-28
Added contextmanager for handling tmp dir when creating gifs

JanEricNitschke:
I created a little contextmanager that handles creation and deletion of the temp dir when plotting gifs.

The changes i made represent two functional changes too.

The first is that previously we always deleted the tmp dir first if it existed. Now we (os.mkdir) raise an exception. I think that is the better behaviour because we do not want to accidentally delete a users folder with their stuff in it. And now it shouldnt be an issue that it might already exist because we always clean up after ourselves with the decorator.

It is also a functional change for the mapcontrol gif as that did not delete the tmp dir previously.
I adjusted the tests to pass the new intendet functionality.

I am also open to the location of the new helper function (directly in plot.py, or should it go to vis __init__)
and the name.


pnxenopoulos (approved):
This looks good to me. Just to confirm -- the prior behavior was that the temp dir was _always_ deleted?

Also, I think given the function's general nature, we could even put it in a higher level of awpy. Does it only have application in the vis module?

JanEricNitschke:
The function currently has exactly two use cases in the vis module. Plot round and  plot mapcontrol. But it is in principle helpful everywhere where we might end up with a temp folder we want to delete at the end.

The two use cases now actually had different behaviour before.

Both would always delete the folder at the start:
https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L267
https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L522

and plot round would also always delete the folder at the end, while plot mapcontrol would leave the temp folder around.
https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L290

Now the behaviour would be that we never delete at the start but always delete at the end.

pnxenopoulos:
> The function currently has exactly two use cases in the vis module. Plot round and plot mapcontrol. But it is in principle helpful everywhere where we might end up with a temp folder we want to delete at the end.
> 
> The two use cases now actually had different behaviour before.
> 
> Both would always delete the folder at the start: https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L267 https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L522
> 
> and plot round would also always delete the folder at the end, while plot mapcontrol would leave the temp folder around. https://github.com/pnxenopoulos/awpy/pull/268/files#diff-0c9028cb3c09bfd6aea0bb1f04ef0762cc02977c2881ffc8af1e4a59c93d7316L290
> 
> Now the behaviour would be that we never delete at the start but always delete at the end.

Excellent. I think this makes sense. Merging now.

PR #270, opened 2023-07-28
Changed PlayerPosition to tuple[float, float, float]

JanEricNitschke:
Changed PlayerPosition to be a tuple of three floats. Adjusted `find_closest_area` by also adding a PlayerPosition2D that is a tuple of 2 floats and adding the relevant overloads. Also fixed all pyright errors that stemmed from this.


pnxenopoulos (approved):
These all look good to me, thanks!

Issue #240, opened 2023-05-26
Incorporate new types from Map Control PR into rest of awpy package


adisujithkumar:
The [map control PR](https://github.com/pnxenopoulos/awpy/pull/235) introduces a few new types to awpy. However, they are not integrated into the rest of the package (nav, demoparser, etc.). This issue is meant to track the work to merge these new types into the rest of the package in the future. 

Running list of types to integrate:
TileId - Could be incorporated into nav, maybe demoparser
TileDistanceObject - Could be combine with TileDistance so there is one overall type that handles all functionalities needed
TileNeighbors - Could be merged into nav as part of a  Nav Tile object/TypeHint re-work overall, but might be optional

JanEricNitschke:
I am not sure how to best combine the TileDistanceObject with the "normal" DistanceObject. 

`DistanceObject` holds the distance and path between two tiles. 

`TileDistanceObject` holds the distance to an unspecified source tile as well as the tileid of the current tile if i understood it correctly.

My first idea would be to remove `TileDistanceObject` and instead use distanceobject, specify some distance type (is it euclidean?) and just have the path either be length 1 with just the current tile or length 2 with the source tile and the current tile.

JanEricNitschke:
On second thought that actually sounds perfect. We could literally just change `area_distance` to have source and destination ids as the path and then we have it already. Then you wouldnt need to wrap the result in a TileDistanceObject to begin with.

PR #273, opened 2023-07-28
Propagate map control types throughout awp

JanEricNitschke:
First commit propagates `TileId`. Please check that i didnt put it somewhere where it doesnt belong and that i didnt miss any where it does.

closes https://github.com/pnxenopoulos/awpy/issues/240

### DIFF
--- a/awpy/analytics/map_control.py
+++ b/awpy/analytics/map_control.py
@@ -21,6 +21,7 @@ from awpy.analytics.nav import (
 from awpy.data import NAV, NAV_GRAPHS
 from awpy.types import (
     BFSTileData,
+    DistanceObject,
     FrameMapControlValues,
     FrameTeamMetadata,
     GameFrame,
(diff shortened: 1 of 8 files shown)


JanEricNitschke:
I removed the TileDistanceObject now and instead made a very minor change to DistanceObject and use that instead.

And i noticed that DistanceObject could probably be replace with a dataclass instead of a typed dict. I think there are a bunch of opportunities for that. But that is pretty decent work and would also be a very breaking change as a couple of user facing functions return such typed dicts that would be better suited as dataclasses.

JanEricNitschke:
I now also completely removed `TileNeighbors` as i switched from manually building the neighbors dict to just directly using `G.neighbors()`.

@adisujithkumar could you double check that this does not break anything? Tests pass but i dont know if every case is covered.

adisujithkumar:
@JanEricNitschke Just pulled your branch and ran the MC notebook and everything ran as expected. Since tests are also passing, I think this PR looks good

adisujithkumar (approved):

adisujithkumar (dismissed):

pnxenopoulos (changes requested):
LGTM. After you fix the minor conflicts then I'll merge

Issue #220, opened 2023-02-19
esea pro demo halftime round clearing issue


ozanhagen:
https://www.hltv.org/matches/2353775/party-astronauts-vs-extra-salt-esea-premier-season-39-north-america
map: inferno

after the first half, on esea there is usually some additional warmup until the players ready up
seems like the kills in that warmup did not get cleared? 

seems like there are 31 rounds logged, while there were only 28 played.

for example player "floppy" killing player "Infinite" in tick 262075. 
the total kill sum has 7 more kills than what the hltv stats show.

pnxenopoulos:
did you try parsing with cleaning (`.parse(clean=True)`) and see if the warmups don't get cleaned? Probably will have to revisit the cleaning method.

PR #275, opened 2023-07-29
Fixed issue with removing rounds with bad scores.

JanEricNitschke:
Closes https://github.com/pnxenopoulos/awpy/issues/220

The issue was that when checking for rounds with bad scores we have logic to allow the last valid round to remain. There is special case logic for OT which was missing a check that the score was ACTUALLY in OT.

### DIFF
--- a/awpy/parser/demoparser.py
+++ b/awpy/parser/demoparser.py
@@ -62,8 +62,6 @@ class DemoParser:
         demo_id (string): A unique demo name/game id.
             Default is inferred from demofile name
         output_file (str): The output file name. Default is 'demoid'+".json"
-        log (bool): A boolean indicating if the log should print to stdout.
-            Default is False
         parse_rate (int, optional): One of 128, 64, 32, 16, 8, 4, 2, or 1.
             The lower the value, the more frames are collected.
             Indicates spacing between parsed demo frames in ticks. Default is 128.
@@ -96,6 +94,7 @@ class DemoParser:
         outpath: str | None = None,
         demo_id: str | None = None,
         log: bool = False,
+        debug: bool = False,
         **parser_args: Unpack[ParserArgs],
     ) -> None:
         """Instatiate a DemoParser.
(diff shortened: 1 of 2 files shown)


pnxenopoulos (approved):
Great fix, thanks Jan-Eric

just to confirm, what would happen pre-fix is that at the end of `test_remove_bad_scoring_no_ot_ot_value` (also is there redundancy in the name there?), the 0-0 -> 0-1 round would still exist in the parsed JSON?

JanEricNitschke:
Main issue was the general logic to check if the score was an OT value just did some math that checks the distance to 15 but did not make sure that the distance was positive. But i think it actually also just straight up messed up he math. Ill have another look.

JanEricNitschke:
And yeah im not sure how to name it properly. The test is to make sure that the OT check does not trigger on non-OT values.

awpy/parser/demoparser.py
```diff
@@ -1023,17 +1028,23 @@ def _has_winner_and_not_winner(self, game_round: GameRound) -> bool:
...
+        # OT draw scores are of the type
+        # 15 + 3xN with N a natural number(1, 2, 3, ...)
+        # So 18, 21, 24, 27
+        # Wins are 1 higher
+        # So 19, 22, 25, 28
+        # So if you substract 15 + 1 from an OT winning round
+        # the number is divisible by 3
+        ot_valid_ct_win = (
+            (game_round["endCTScore"] - tie_score - 1) % ot_tie_score == 0
+            and game_round["endTScore"] < (game_round["endCTScore"] - 1)
```
adisujithkumar, line 1040: Would it help to have a comment explaining why the difference has to be at least 2 because if one team reaches 18 then either they will reach 19 and win by at least two or the other team will also reach 18 and restart the overtime process essentially? Took me a second to realize why this section was needed initally and a small comment could help with that for others in the future if they have the same slight confusion

JanEricNitschke: Added one.

pnxenopoulos (approved):

hojlund123 (approved):

Issue #189, opened 2022-09-04
Error while parsing a demo in two parts


kenmareirl:
**Describe the bug**
When a demo is splitted in two parts, an error is raised while parsing the first one.

**Include Information to Reproduce**

- Download [this demo](https://www.hltv.org/matches/2357663/liquid-vs-ninjas-in-pyjamas-blast-premier-fall-groups-2022)
- Try to parse overpass p1
- The error is raised (see screenshots)

**Screenshots**
![parse_part_1](https://user-images.githubusercontent.com/92644057/188329413-1e408308-e338-47bc-921f-502d18a95a27.png)

**Additional context**
Even with this error a json is generated.

pnxenopoulos:
Thanks for finding this @kenmareirl. This may require some sort of additional function.

kenmareirl:
you're welcome 👍

JanEricNitschke:
I don't know if this is something we need to to address at all. This error is raised directly from the go parser and we panic on it at the very end.

I guess it would probably make more sense to print a warning here that the demo ended unexpectedly and to double check the resulting json file.

PR #276, opened 2023-07-29
Handle ErrUnexpectedEndOfDemo gracefully.

JanEricNitschke:
Closes https://github.com/pnxenopoulos/awpy/issues/189

We now just print a warning that the error occured and to check the result if we get a `ErrUnexpectedEndOfDemo` when parsing a demo.

Also did a minor cleanup and fixed the gitignore when it comes to demo files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,7 +135,7 @@ dmypy.json
 .pytype/
 
 # csgo demofiles
-.dem
+*.dem
 
 # vscode
 *code-workspace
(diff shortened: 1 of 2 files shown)


pnxenopoulos (approved):
Guessing the changes in `parse_demo.go` seem like fixing redundant logic on my part, yea?

JanEricNitschke:
Yeah, the else branch was doing something that was also done at the end of the if branch. So i just moved that out of the if/else and deleted the else branch.

Issue #196, opened 2022-10-31
Missing round after parsing a demo


kenmareirl:
**Describe the bug**
For the match MIBR - FLUXO:

- Round n°14 is 11-3 for T at the end of the round 
- Round n°15 is starting at 12-3. 
- A round has been cleaned from 11-3 to 12-3

**Include Information to Reproduce**
Parse [this match](https://www.hltv.org/matches/2359393/mibr-vs-fluxo-blast-premier-fall-showdown-2022-north-america) and check vertigo (map 2)

**Screenshots**
N/A

**Additional context**
N/A

JanEricNitschke:
```
  "roundNum": 17,
   "isWarmup": false,
   "startTick": 258020,
   "freezeTimeEndTick": 0,
   "endTick": 275572,
   "endOfficialTick": 276468,
   "bombPlantTick": 272588,
   "tScore": 11,
   "ctScore": 3,
   "endTScore": 12,
   "endCTScore": 3,
   "ctTeam": "FLUXO",
   "tTeam": "MIBR",
   "winningSide": "T",
   "winningTeam": "MIBR",
   "losingTeam": "FLUXO",
   "roundEndReason": "TerroristsWin",
   "ctFreezeTimeEndEqVal": 0,
   "ctRoundStartEqVal": 0,
   "ctRoundSpendMoney": 0,
   "ctBuyType": "Full Eco",
   "tFreezeTimeEndEqVal": 0,
   "tRoundStartEqVal": 0,
   "tRoundSpendMoney": 0,
   "tBuyType": "Full Eco",

```

it is this round that gets cleaned away because it it all messed up in the original json, You can see the freezeTimeEndTick being 0, all the money values are 0.

Also the kills have weird times and the frames list is empty.

Looking at the demo both the round itself as well as the one prior seemed to be completely normal. Only after the problematic round was there a longer pause which is correctly shown in the next round by looking at the start and freezetimeend ticks.

So this is most likely a problem in the go part of the parser and not the python cleaning afterwards.

JanEricNitschke:
Ok, so i found out that the issue is that the `RoundFreezetimeEnd` event does not fire. This means the parser thinks we are still in freezetime and this does not parse the equipment value and also does not parse any frames.

And we have a cleaning function that removes rounds without frames as those almost always indicate an issue with the round.

JanEricNitschke:
[...]

PR #278, opened 2023-07-29
Added check into frames event to correct for skipped FreezeTimeEnd events

JanEricNitschke:
Closes https://github.com/pnxenopoulos/awpy/issues/196

It seems that the game rules contains a property tracking freezetime status that is more accurate than the freezetimeend event.

I have added a check into the framesdone event that fixes the `roundinfreezetime` toggle when the property says something different.

One could potentially replace that flag and event with this property entirely. But it isnt 100% trivial as it can be nil in some cases. So going with this for now.


pnxenopoulos (approved):
Looks good to me -- what led you to believe that this has better stability? Personal testing?

JanEricNitschke:
People in the demoinfocs discord.

From "Dr. Freeman":
> This particular event is unreliable.

From [LaihoE](https://github.com/LaihoE/demoparser)
> The entity: "CSGameRulesProxy" has a field: "m_bFreezePeriod", but not sure how to access it
> I'm quite sure that is the "correct way to get it" ie how the client gets it

JanEricNitschke:
And the one case i tested from the issue seems to confirm it. Maybe we want to add that demo to our testing set?

pnxenopoulos (approved):

Issue #280, opened 2023-07-30
Have plot_map_control_metrics return figure object


adisujithkumar:
Slight change to have plot_map_control_metrics return the figure object instead of using plt.show() within the function. Returning the figure object would align the function similarly to the rest of the plot.py file (which usually returns the figure, axes tuple back to the user). If the user is able to have the figure returned back, they can do whatever they like (use tight layout, save to file, custom annotations, etc.). If others think this change would be beneficial - I'll send a PR for this issue by the end of the weekend.

pnxenopoulos:
Closing for Awpy 2.0

PR #282, opened 2023-07-30
Changing plot_map_control_metrics so it returns a matplotlib figure, axes object

adisujithkumar:
Addressing a [GitHub issue](https://github.com/pnxenopoulos/awpy/issues/280) I brought up yesterday where I believe it would be better for the user if plot_map_control_metrics returned a fig, axes tuple instead of doing plt.show() within the function. This allows the user to interact/customize the plot to their specific needs. Also updated the Map Control notebook to reflect this change.


pnxenopoulos (approved):
Looks good. I think this functionality is the one we should have

Issue #284, opened 2023-08-27
CEVO POV demo parses with empty gameRounds


JacobPalecki:
**Describe the bug**
POV demo from CEVO match in 2017 parses with empty gameRounds.

**Include Information to Reproduce**
Link to demo: [https://www.dropbox.com/scl/fi/51t9cnhflwz07hs30oxz3/gt.dem?rlkey=hymp1g8nj1qo37jdugwrarxkp&dl=0](https://www.dropbox.com/scl/fi/51t9cnhflwz07hs30oxz3/gt.dem?rlkey=hymp1g8nj1qo37jdugwrarxkp&dl=0)
Code:
`
from awpy.parser import DemoParser

demo_parser = DemoParser(
    demofile = "gt.dem", 
    demo_id = "gt", 
    parse_rate=128, 
    trade_time=5, 
    buy_style="hltv",
    clean=False
)

data = demo_parser.parse()
`

**Screenshots**
![image](https://github.com/pnxenopoulos/awpy/assets/16246195/9a3f240f-07a9-437f-a0b7-6174de2d604a)

**Additional context**
On latest awpy 1.3.1 and python 3.11.5

JanEricNitschke:
Hi, thanks for the issue.

POV demos are a bit difficult for awpy as they are missing a bunch of information. Specifically they are basically always missing the opposing team and sometimes even general information or events.

That can cause issues with the parsing logic but more frequently is causes issue with the cleaning of rounds that throws away warmups, knife rounds and such.

If you run it with `data = demo_parser.parse(clean=False)` instead then you will get the rounds.

This means that the cleaning throws away all of the rounds. I will check if there is something reasonable we can do that avoids this issue, however i can not guarentee that because this is a heuristic thing anyway and POV demos are in general a small use case.

If there is nothing general we can/will do you can try to customize the cleaning process by running

```
demo_parser.parse(clean=False)
data = demo_parser.clean_rounds(remove_bad_scoring=False, save_to_json=True)
```

And you can check [here](https://github.com/pnxenopoulos/awpy/blob/main/awpy/parser/demoparser.py#L851) what cleaning steps are available. If that does not work you would have to implement your own cleaning logic.

JanEricNitschke:
Ok, so it seems that the issue is actually just due to the warmup cleaning logic. That seems to remove basically all rounds.

Will investigate this further but i feel that this is fixable.

pnxenopoulos:
[...]

PR #286, opened 2023-08-29
Round cleaning and freezetime recognition improvements.

JanEricNitschke:
Closes https://github.com/pnxenopoulos/awpy/issues/284

This PR includes a reduction of `remove_warmups` to it simplest form due to it seeming too rigid and functionally unclear.
It now only removes rounds tagged as `isWarmup`.

This was the main reason the demo in the issue had all of its rounds removed in cleaning.

---

I also adjusted `remove_bad_scoring` slightly so that it now requires rounds that are kept because they have a winning score to also have to have a higher score than the previous round.

In the cevo demo there were multiple rounds before the "last" that had the final game score.

16-5 (actual last round)
16-5 (some left over non gameplay round)
0-0 (reset round before the server closed)

As this check was not present previously this caused the "bad" second to last round to be wrongly kept. This has been fixed now.

---

Lastly the PR includes a change to freezetime/roundstart handling.

The `FreezeTimeEndHandler` handler includes a check whether the game is already (still) in freezetime to deal with cases where the `RoundStart` event did not fire. It subsequently adds the currently still  kept alive round to the list of game rounds and starts the new one from the freeze time end instead of the ideal round start.

The addition of the gamestate check for the freeze time which was added in `https://github.com/pnxenopoulos/awpy/pull/278` to handle cases where the `FrezeTimeEnd` event was missing. 

This introduced an issue when the gamestate changed before the event triggered. Causing the `roundInFreezetime` flag to be set before the event triggered which caused the event to add a lot short dummy rounds into the list of rounds.

I have now added an additional flag which indicates if the `roundInFreezetime` flag was set from the gamestate checker. If that is the case then the freezetime event handler does not add its own round.

These two flags could potentially be refactored into a single enum though. Dont know if you feel that that is necessary.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,7 @@ repos:
         entry: pyright
         language: system
         types: [python]
+        pass_filenames: false
   - repo: local
     hooks:
       - id: pylint
@@ -56,6 +57,7 @@ repos:
         entry: pylint awpy
         language: system
         types: [python]
+        pass_filenames: false
         args:
           [
             "-rn", # Only display messages
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,6 @@ repos:
         entry: pyright
         language: system
         types: [python]
-        pass_filenames: false
   - repo: local
     hooks:
       - id: pylint
@@ -57,7 +56,6 @@ repos:
         entry: pylint awpy
         language: system
         types: [python]
-        pass_filenames: false
         args:
           [
             "-rn", # Only display messages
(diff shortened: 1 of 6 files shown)


pnxenopoulos (dismissed):

pnxenopoulos:
Mostly looks great, but want to hold off on merging just to get some clarification on the following: do we have a demo where the behavior described happens (might already exist in the test suite)?

JanEricNitschke:
So the issue i was trying to fix was https://github.com/pnxenopoulos/awpy/issues/284 which also has a link to the demo which we could add to the testing set.

The cause of that issue was that the warmup cleaning was removing all rounds.

Debugging that i also noticed that a bunch of dummy rounds got added (and cleaned) due to the freezetime changes.
While they were being cleaned they caused the roundstart time to be cut off (because the freezetime fallback only starts the rounds at the freezetime end).

So we could add that demo to the testing set and do a check on round numbers uncleaned and cleaned to verify that both these changes here work properly.

pnxenopoulos (approved):

Issue #285, opened 2023-08-28
Change ParseRate logic


JanEricNitschke:
Currently parse rate sets the fraction of recorded frames that makes it into the json.

A parse rate of 1 means every single frame makes it into the json while 128 means every 128th frame makes it in.

This has the disadvantage that it leads to fluctuating frames per second depending on the demo tickrate. If one parses demos from different sources some demos will have 1 frame per second while other would only have one frame every 4 or so seconds.

We could change (or offer an alternative) the logic the specify the (maximum) number of frames per second instead.
That way one could specify 1 frame per second and would always get a consistent value (as long as the demo permits).

Setting values of more frames per second than what was recorded in the demo would emit a warning and instead just record every single frame instead as a best effort.

What are your thoughts on this?

pnxenopoulos:
Closing for Awpy 2.0

PR #287, opened 2023-08-30
DRAFT: Added option to treat parserate as relative to 128 demo tick rate.

JanEricNitschke:
Closes https://github.com/pnxenopoulos/awpy/issues/285

Reference to how this could work. No tests written yet.


pnxenopoulos:
Moved to demoparser backend in v2. Stay tuned for v2 release in a week or two. 

New `awpy.demo.Demo` class has a tickrate parameter you can create on construction.

PR #296, opened 2023-10-03
Update ci

JanEricNitschke:
Updating linters and adding python 3.12 to workflow to see if everything works.


JanEricNitschke:
3.12 seems to crash because of the removal of distutils.

https://docs.python.org/3.12/whatsnew/3.12.html#distutils

JanEricNitschke:
Seem to be due to numpy not supporting 3.12 until 1.26.0 which is currently excluded from our requirements.

https://numpy.org/devdocs/release/1.26.0-notes.html

JanEricNitschke:
I have updated the numpy dependency, maybe that now makes it work with 3.12.

I also added [pymend](https://pypi.org/project/pymend/) to the pre-commit configuration. Left me know what you think.

JanEricNitschke:
Also i feel that our current setup with the installation is a bit messy. We have the requirements.txt, setup.py and pyproject.toml and all of them specify the dependencies.

JanEricNitschke:
Updated all dependencies to the newest versions on pypi. Lets see if this builds and if all tests still pass.

JanEricNitschke:
3.12 issue seem to now be shapely which does not have a wheel for 3.12 yet on pypi.

pnxenopoulos:
Totally changed everything for v2. If you have any suggestions, fire away either in a PR, comments or Discord, I trust your judgement :)

PR #300, opened 2023-10-20
Make plot rounds respect levels

JanEricNitschke:
Previously plot rounds showed the two levels for maps that have them but only plotted the player and bomb in the upper half. Now it should also plot them in the lower one if they are there.

### DIFF
--- a/awpy/types.py
+++ b/awpy/types.py
@@ -10,7 +10,7 @@ from typing_extensions import TypedDict
 class PlotPosition:
     """Class to store information needed for plotting a position."""
 
-    position: tuple[float, float]
+    position: tuple[float, float, float]
     color: str
     marker: str
     alpha: float | None = None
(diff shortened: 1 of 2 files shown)


pnxenopoulos (dismissed):
looks great. i think this addresses the question in Discord

JanEricNitschke:
I would still like to figure out why the levels half the width instead of doubling the height.

pnxenopoulos:
> I would still like to figure out why the levels half the width instead of doubling the height.

All good, in no rush to merge

JanEricNitschke:
Shoud be all good now. But note that there are breaking changes in here to plot_positions.

pnxenopoulos:
Currently adding this in v2. Will be referencing. I found in the data the z-level cutoffs fortunately.

PR #306, opened 2024-01-28
Fix issue with missing round_end events causing error

billfreeman44:
Fixes `ValueError: cannot convert NA to integer` bug when LO3 happens in a server.

### DIFF
--- a/awpy/parser/round.py
+++ b/awpy/parser/round.py
@@ -22,7 +22,9 @@ def parse_rounds(parsed_round_events: list[tuple]) -> pd.DataFrame:
     round_events = process_round_events(parsed_round_events)
     round_event_df = pd.concat(round_events)
 
-    return create_round_df(round_event_df)
+    round_df = create_round_df(round_event_df)
+    round_df=round_df.loc[round_df.index>max(round_df[round_df['round_end'].isna()].index)]
+    return round_df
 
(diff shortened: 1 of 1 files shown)


billfreeman44:
Screenshot 2024-01-27 213603
Screenshot 2024-01-27 213609

Logic here is to avoid having "round 2" in the screenshots here as a valid round. it is before the LO3 finishes. The best logic is just "toss everything before the first valid round" IMO.

billfreeman44:
I get this error when testing `PydanticUserError: `@validator` cannot be applied to instance methods` though that particular error seems to happen without this change. I was on 433eeaa and an equivalent line there worked no problem.

pnxenopoulos:
Thanks for finding this. I fixed the pydantic error you mentioned.

How do we determine a valid round? I suppose a valid round needs a round_end_official. But the last round of an match could have the same round_end and round_end_official. I guess if there are rounds after, that is our tell?

billfreeman44:
it is difficult. maybe this package could extract everything and then the user could use some kind of logic to throw away rounds? probably not ideal but i don't know.

pnxenopoulos:
Believe I got a way to determine valid rounds in the `awpy2` branch now, might require some more testing across more demos

PR #327, opened 2024-04-09
add win probability feature to awpy2

rorybush:


awpy/stats/win_prob.py
```diff
@@ -0,0 +1,87 @@
...
+    Calculate win probabilities for specified ticks in a CS2 match demo.
+
+    Args:
+        demo (Demo): A parsed demo object containing CS2 match data.
+        ticks (Union[int, List[int]]): A single tick or a list of ticks at which to calculate win probabilities.
+
+    Returns:
+        List[Dict[str, float]]: A list of dictionaries with the calculated win probabilities for CT and T sides for each tick.
+
+    """
```
pnxenopoulos, line 85: to avoid errors here, you can have it `raise NotImplementedError` for now.

awpy/stats/win_prob.py
```diff
@@ -0,0 +1,87 @@
...
+    
+    Raises:
+        ValueError: If specified ticks are not found within the demo object.
+    """
+    if not demo.ticks.empty:
+        if isinstance(ticks, int):
+            ticks = [ticks]
+        
+        # Gather round numbers for each specified tick
+        round_numbers = []
```
pnxenopoulos, line 27: I think this whole block could be improved -- we could simply filter `demo.ticks` where the `.tick` is in `ticks`, then take the unique values of the round number. This way, we avoid a for loop.

```
all_ticks = demo.ticks[demo.ticks["tick"].isin(ticks)]
round_numbers = demo.ticks.round_number.unique()
```

awpy/stats/win_prob.py
```diff
@@ -0,0 +1,87 @@
...
+        for tick in ticks:
+            tick_data = demo.ticks[demo.ticks["tick"] == tick]
+            if tick_data.empty:
+                raise ValueError(f"No data found for tick {tick}.")
+            round_numbers.append(tick_data["round"].iloc[0])
+        
+        # Check if all ticks belong to the same round
+        if len(set(round_numbers)) > 1:
```
pnxenopoulos, line 35: This is an interesting case. Should we throw an error? Nothing prevents us from predicting, so I would argue we don't even need to care about round numbers (also means we could axe the above code).

awpy/stats/win_prob.py
```diff
@@ -0,0 +1,87 @@
...
+            raise ValueError("Specified ticks are from different rounds. Please ensure all ticks belong to the same round.")
+            
+        game_state = []
+
+        # Process Game State/Feature Matrix
+        for tick in ticks:
```
pnxenopoulos, line 41: The logic here makes sense, what I think we want to do though is create a smaller function then use `.apply()` in pandas. Let's refactor along those lines.

rorybush: I have updated the function to use `.apply()`

awpy/stats/win_prob.py
rorybush: I have removed this

awpy/stats/win_prob.py
rorybush: I have refactored the block to avoid using a for loop

awpy/stats/win_prob.py
rorybush: I have implemented `raise NotImplementedError` to avoid any errors.

tests/test_win_prob.py
```diff
@@ -0,0 +1,49 @@
...
+from awpy.stats import win_probability
+
+class TestWinProbability:
+    """Tests the win probability calculations.
+    
+    https://www.hltv.org/matches/2369248/natus-vincere-vs-virtuspro-pgl-cs2-major-copenhagen-2024-europe-rmr-closed-qualifier-a
+    """
+
+    @classmethod
+    def setup_class(cls):
```
pnxenopoulos, line 15: We probably want to use something like `@pytest.fixture(scope="class")` decorators as shown in `test_parsers.py`

tests/test_win_prob.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Setup method called before any tests are run.
+        Initializes the Demo object and random weights for testing.
+        """
+        cls.demo = Demo(file='tests/natus-vincere-vs-virtus-pro-m1-overpass.dem')  
+
+
+    @classmethod
+    def teardown_class(cls):
```
pnxenopoulos, line 24: In which case, we won't need a teardown method (nor a setup)

tests/test_win_prob.py
```diff
@@ -0,0 +1,49 @@
...
+        Teardown method called after all tests have run.
+        """
+        cls.demo = None
+
+
+    def test_win_probability_symmetry(self):
+        """Test to ensure P(CT Win) = 1 - P(T Win) at a given tick."""
+        probabilities = win_probability(self.demo, 168200)
+        for prob in probabilities:
+            assert prob["CT_win_probability"] == pytest.approx(1 - prob["T_win_probability"], 0.01)
```
pnxenopoulos, line 35: I would probably assert that CT win prob + T win prob = 1.0

We might also want a stronger tolerance than 0.01. Probably the default is fine (so just `pytest.approx(1.0)`)

tests/test_win_prob.py
pnxenopoulos: We could use an existing demo, as well

tests/test_win_prob.py
rorybush: removed teardown method and setup and replaced with pytest fixtures

tests/test_win_prob.py
rorybush: implemented pytest fixtures and now use existing demo

tests/test_win_prob.py
rorybush: cleaned up code and now use a stronger tolerance.

PR #343, opened 2024-06-23
Updating pre-commit and running pylint/pyright

JanEricNitschke:


pnxenopoulos:
Managing our tooling as always :) 

Ping me when you want this in, since it looks like you're still committing

JanEricNitschke:
Will do :D

I want to try to enable pyright strict mode and see if i can find/create stubs for demoparser2.

Depending on how complicated that is i would either merge this and do that in a separate step or include it here.

pnxenopoulos:
Closing since I want to move towards `uv`'s type-checker when it gets released. Usually their beta stuff is good enough for our use case.

PR #344, opened 2024-06-23
v2.0.0b2

pnxenopoulos:


JanEricNitschke:
You could already update the return type hint. Otherwise i will do that in my MR after i rebase.

pnxenopoulos:
> You could already update the return type hint. Otherwise i will do that in my MR after i rebase.

Pushed in most recent. Working on a basic visibility function so the branch is in flux

PR #347, opened 2024-07-22
v2.0.0b3

pnxenopoulos:


pnxenopoulos:
Notes -- this will remove 3.9 and 3.12 will depend on an upcoming usd-core release (24.08)

PR #353, opened 2024-10-28
Fixed plot bugs, added functionality, updated images

ventsiR:
Added functions in awpy.plot.utils that transform in the opposite direction (i.e. from map pixel coordinates to in-game coordinates) and added documentation indicating direction of transform.

Updated the vertigo map images to show the new layout after the map got updated.


awpy/plot/plot.py
```diff
@@ -40,6 +46,10 @@ def plot(  # noqa: PLR0915
             - 'armor': int (0-100)
             - 'direction': Tuple[float, float] (pitch, yaw in degrees)
             - 'label': str (optional)
+        ignore_extreme_points (bool, optional): If set to True, will ignore
```
pnxenopoulos, line 49: This is an interesting proposition. Is there any reason we would want to allow this in the first place? Or, in other words, do you think there's any danger to making this a default?

awpy/plot/plot.py
```diff
@@ -22,14 +23,19 @@
 def plot(  # noqa: PLR0915
     map_name: str,
     points: Optional[List[Tuple[float, float, float]]] = None,
+    is_lower: Optional[bool] = False,
```
pnxenopoulos, line 26: What if this parameter were reworked into a "lower fraction"? As in, we assume that upper points are plotted with alpha = 1.0, and then lower are lower_frac*alpha ? This would allow for a more expressive form of what it appears you are going for.

awpy/data/map_data.py
```diff
@@ -8,119 +8,110 @@
         "scale": 2.539062,
         "rotate": 1,
         "zoom": 1.3,
-        "selections": [
-            {"name": "default", "altitude_max": 10000, "altitude_min": -5},
-            {"name": "lower", "altitude_max": -5, "altitude_min": -10000},
-        ],
+        "lower_level_max": -5.0,
```
pnxenopoulos, line 11: I like this simplification, but I may offer a rewording -- perhaps `lower_level_max_units` or something. Just to put the "units" into the name (which unfortunately in this case is just generic "units").

awpy/plot/plot.py
pnxenopoulos: Which, we may also want to offer a param for the "dead player alpha" that we default to 0.15, too

awpy/plot/plot.py
```diff
@@ -271,11 +340,21 @@ def heatmap(
...
+        ignore_extreme_points (bool, optional): If set to True, will ignore
+            points that are outside of the map graphic. If set to False, will
+            draw those points, changing the resolution of the ouput image.
+            Defaults to False.
         size (int, optional): Size of the heatmap grid. Defaults to 10.
         cmap (str, optional): Colormap to use. Defaults to 'RdYlGn'.
         alpha (float, optional): Transparency of the heatmap. Defaults to 0.5.
         vary_alpha (bool, optional): Vary the alpha based on the density. Defaults
             to False.
+        vary_alpha_range (List[float, float], optional): The min and max transparency
```
pnxenopoulos, line 355: I'm a bit unclear as to what this parameter is doing -- do you mind explaining again? I realize I don't quite remember what I wrote 🤣

awpy/plot/utils.py
```diff
@@ -34,10 +34,40 @@ def position_transform_axis(
     return (start - position) / scale
 
 
+def position_revert_axis(
```
pnxenopoulos, line 37: What's the intended use for this function?

pnxenopoulos (commented):
I really appreciate the PR! There are some interesting and helpful additions here. I think the PR looks mostly good and I left some more clarification/discussion-based comments. Once we sort those minor comments out, we can merge this into the `2.0.0b5` branch. I should just make it 2.0, but alas...

awpy/data/map_data.py
ventsiR: Sounds good!

Sidebar: I am a bit new to GitHub & commits; is it good etiquette for me to implement your feedback, or to let you implement it?

awpy/plot/plot.py
ventsiR: Good idea!

The only thing to note is that when `is_lower = True`, it also changes the image to the lower part of the map, but then again, you can control this via the `map_name` argument.

awpy/plot/plot.py
ventsiR: If I were designing this function from the ground-up, I would have the default set to `True`.

In the commit I have set it to default as `False` because it mimics the previous behavior of the function (i.e. it plotted points way outside of the map). My reasoning was backwards compatibility for people who have already written code with the current behavior in mind.

I'm kind of new to GitHub and commits so I'm leaving the decision up to you :D

awpy/plot/plot.py
ventsiR: Previously (before I added `vary_alpha_range`) when `vary_alpha` was set to `True`, it would plot the densest points with alpha equal to the `alpha` argument and the least dense points with `alpha = 0`, and anything in between would be assigned linearly between `0` and `alpha`.

`vary_alpha_range` lets the user set a custom range (instead of `0` and `alpha`).

awpy/plot/utils.py
ventsiR: Given a point that is plotted on an image, users can feed the pixel co-ordinates of that point into `position_revert_axis()` and get the in-game co-ordinates of that point.

I am using your library to plot points on a matplotlib graph, draw a polygon over it, and return my polygon to a database in terms of in-game coordinates.

awpy/plot/plot.py
ventsiR: I now realize that `vary_alpha_range` might be a confusing name - it was intended to sound like "the range of argument `vary_alpha`", but it sounds like, "vary the alpha's range".

Can't think of anything better though - I considered `alpha_range`, but it's even more ambiguous :D

awpy/plot/plot.py
pnxenopoulos: Totally understand the backwards compatibility concerns. I think given that this is still a beta version, we can feel ok with breaking the previous functionality, particularly in this case, where there is little noticeable value to plotting points outside the map. I say we remove this arg and make it default behavior! It's a nice safeguard that I overlooked when first writing the function.

awpy/data/map_data.py
pnxenopoulos: I think it depends from person-to-person, repo-to-repo, but _usually_ the one who opened the PR will make the changes, unless it's a super small change.

awpy/plot/plot.py
pnxenopoulos: I see. If an alpha range is specified, should we assume the user wants to vary_alpha? This may allow us to remove one arg. If so, then I think having `alpha_range` is a fine name, as long you add just a tiny bit more explanation in the docstring!

awpy/plot/utils.py
pnxenopoulos: I see. Perhaps what we can do is rename the function. Usually for these transformer-style function we can name them like "X_to_Y". We probably should rename the `position_transform()` function, we can still do that, I'm all ears for names!

pnxenopoulos:
Also, DM me on Discord so I know which account you are!

awpy/plot/utils.py
ventsiR: I agree; I would name the functions `in-game_to_pixel_transform` and `pixel_to_in-game_transform`, but I didn't take the liberty due to - again - backwards compatibility reasons.

Maybe rename the functions to `in-game_to_pixel_transform` & `pixel_to_in-game_transform` and create "new" functions  that go by the old names (`position_transform_axis` & `position_revert_axis`) that simply call `in-game_to_pixel_transform` & `pixel_to_in-game_transform` and send a deprecation warning in the console?

awpy/plot/plot.py
ventsiR: Sounds good, will remove the arg and integrate the functionality.

awpy/data/map_data.py
ventsiR: Sounds good. Will implement everything you've mentioned; they're pretty small changes and everything is still relatively fresh in my mind :D

awpy/plot/plot.py
ventsiR: Great idea!

awpy/plot/utils.py
pnxenopoulos: I think we can do the following:

game_to_pixel(...)
pixel_to_game(...)

since "in game" is redundant with "game" and "transform" is implied by the "to"

then, the interface is much cleaner and imo, easy to tell what the functions do

awpy/plot/plot.py
pnxenopoulos: Now that I look at this, perhaps we want to just have a generic `game_to_pixel(tuple)`. That tuple can be an (x,y) pair. What do you think?

awpy/plot/plot.py
```diff
@@ -86,16 +81,14 @@ def plot(  # noqa: PLR0915
...
-            if (ignore_extreme_points
-                and (
-                    transformed_x < 0
-                    or transformed_x > 1024
-                    or transformed_y < 0
-                    or transformed_y > 1024
-                )
+            if (
+                transformed_x < 0
+                or transformed_x > 1024
```
pnxenopoulos, line 89: Because these are constants, it would be helpful to have a comment above here stating that these are the maximum expected bounds

awpy/plot/utils.py
```diff
@@ -115,6 +116,69 @@ def is_position_on_lower_level(
...
+
+def position_transform_axis(
+    map_name: str, position: float, axis: Literal["x", "y"]
+) -> float:
+    """Calls `game_to_pixel_axis` and sends warning.
+    
+    This is the old name of function `game_to_pixel_axis`. Please update
+    your code to avoid future deprecation.
+    """
+    warnings.warn(
```
pnxenopoulos, line 139: Great idea to add the deprecation warnings. I think we can drop the `_renaming_warning()` however, and just include the specific strings in the deprecation messages!

awpy/plot/utils.py
```diff
@@ -115,6 +116,69 @@ def is_position_on_lower_level(
...
+    This is the old name of function `pixel_to_game_axis`. Please update
+    your code to avoid future deprecation.
+    """
+    warnings.warn(
+        _renaming_warning("position_revert_axis()", "pixel_to_game_axis()"),
+        DeprecationWarning)
+    return pixel_to_game_axis(map_name, position, axis)
+
+
+def position_transform(
```
pnxenopoulos, line 159: I also think these functions would still live in `plot`, rather than live in utils (so the import path would be the same)

awpy/plot/utils.py
```diff
@@ -1,15 +1,16 @@
 """Utilities for plotting and visualization."""
 
 from typing import Literal
+import warnings
```
pnxenopoulos, line 4: should generally do 

```
import warnings

from typing import Literal

from awpy.data.map_data import MAP_DATA
```

PR #359, opened 2025-01-12
Add extraction of spawn points.

JanEricNitschke:


awpy/nav.py
```diff
@@ -430,3 +432,120 @@ def to_json(self, path: str | Path) -> None:
         nav_dict = self.to_dict()
         with open(path, "w", encoding="utf-8") as json_file:
             json.dump(nav_dict, json_file)
+
+
+VentsValue = str | int | float | bool | tuple[float, ...]
+
+
+@dataclass
+class Position:
```
pnxenopoulos, line 441: Probably could use `awpy.vector.Vector3` instead of position. Perhaps the naming is not congruent, but the definition is the same

JanEricNitschke: Makes sense, didnt read through the full code and missed this.

PR #360, opened 2025-01-13
Added label options to plot method

smartbackwards:
Added label_direction and label_fontsize as optional parameters in the point_settings dictionary, in the plot method.

label_fontsize specifies the size of the font for the label of a given point (defaults to 6, old hard coded parameter)

label_direction specifies the location of the label in relation to the dot. Defaults to 0 - above the point. 90 would place the label to the right of the point, 180 below, 270 directly to the left etc.

Please note two things: 
- i hardcoded the label_horizontal_offset value. The alternative was to create another pyplot figure, plotting the label there and getting the length from text_object.get_window_extent(renderer=renderer). That seemed like it would slow down the time of execution too much. The hardcoded value works best for labels of length 5-10. 
- Aesthetically it made more sense for me for the label to be on top of the HP/armor bars, instead of enlarging/adjusting the position of the ellipse on which the label is plotted. see pics below: 
- 
image
image

### DIFF
--- a/awpy/plot/plot.py
+++ b/awpy/plot/plot.py
@@ -51,6 +51,8 @@ def plot(  # noqa: PLR0915
             - 'armor': int (0-100)
             - 'direction': tuple[float, float] (pitch, yaw in degrees)
             - 'label': str (optional)
+            - 'label_direction': int (default 0, optional) (location in degrees)
+            - 'label_fontsize': int (default 6, optional)
 
     Raises:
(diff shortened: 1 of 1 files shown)


pnxenopoulos:
This looks like a nice addition and makes me think of an additional improvement -- perhaps we can also guess where to put the name, so that it always is in the map? Perhaps a player can specify defaults and then if those can't fit inside the map, we do our best guess?

PR #361, opened 2025-01-13
add de_train map data

smartbackwards:
Added de_train radar pngs and map data. Not sure if the map data is correct (on test plots the positions seemed always to appear on the lower levels, even out in A yard), here's the CS source code txt file I used:
```

"de_train"
{
    "material"      "overviews/de_train"
    "pos_x"     "-2308"
    "pos_y"     "2078"
    "scale"     "4.082077"
        "verticalsections"
    {
        "default" // use the primary radar image
        {
            "AltitudeMax" "20000"
            "AltitudeMin" "-50"
        }
        "lower" // i.e. de_nuke_lower_radar.dds
        {
            "AltitudeMax" "-50"
            "AltitudeMin" "-5000"
        }
    }
    "CTSpawn_x"     "0.860000"
    "CTSpawn_y"     "0.770000"
    "TSpawn_x"      "0.120000"
    "TSpawn_y"      "0.250000"
    "bombA_x"       "0.630000"
    "bombA_y"       "0.490000"
    "bombB_x"       "0.520000"
    "bombB_y"       "0.760000"
}

```


JanEricNitschke:
Yeah, lower level is everything that is not in the tiny upper sliver in the halls. So it makes sense. For train most of the map is in the lower level, which is a bit different compared to e.g. Nuke or vertigo, where the default is the upper.

PR #362, opened 2025-01-13
Expanded script to directly update map data

JanEricNitschke:


pnxenopoulos:
Leaving open while I prepare infra/scripts/workflow for these artifacts

JanEricNitschke:
I think there are still some changes in here that make sense to integrate. Like the change of the map data to a json and having the powershell scripts automatically move the files into the places where they will likely be used.

awpy/cli.py
```diff
@@ -120,6 +121,16 @@ def parse_spawns(vent_file: Path, *, outpath: Optional[Path] = None) -> None:
     )
 
 
+@awpy.command(help="Parse Counter-Strike 2 map information.")
+@click.argument("overview_folder", type=click.Path(exists=True))
+def parse_overviews(overview_folder: Path) -> None:
+    """Parse a nav file given its path."""
```
pnxenopoulos, line 127: Would change docstring to 

`"""Parse radar overviews."""`

awpy/cli.py
```diff
@@ -120,6 +121,16 @@ def parse_spawns(vent_file: Path, *, outpath: Optional[Path] = None) -> None:
     )
 
 
+@awpy.command(help="Parse Counter-Strike 2 map information.")
+@click.argument("overview_folder", type=click.Path(exists=True))
+def parse_overviews(overview_folder: Path) -> None:
```
pnxenopoulos, line 126: might call this `parse_radars`

awpy/cli.py
```diff
@@ -120,6 +121,16 @@ def parse_spawns(vent_file: Path, *, outpath: Optional[Path] = None) -> None:
     )
 
 
+@awpy.command(help="Parse Counter-Strike 2 map information.")
```
pnxenopoulos, line 124: would use `name="radar"` here so that we can call `awpy radar ...`

pnxenopoulos:
These scripts were super helpful -- we'll keep it manual for now but I have an idea of how we may want to automate this, perhaps in another repo, like an `awpy-data`. We can have that run nightly, pull the latest awpy, create the artifacts, and upload to blob storage. That's the next task.

PR #363, opened 2025-02-20
Creating Nav Jupyter Notebook Example

adisujithkumar:
* Creating nav.ipynb (Nav jupyter notebook example)
* Creating awpy.plot.tile to help visualize anything Nav-related (all tiles on a map, specific tile on map, path between tiles on map, etc.)


awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
+"""
```
pnxenopoulos, line 1: I wouldn't call it `tile_viz_utils` because (1) long-ish name and (2) I try not to use `utils` (even though I use it elsewhere it awpy). What about just `tile.py` or `nav.py`?

awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
...
+import logging
+import os
+from pathlib import Path
+
+import matplotlib.image as mpimg
+import matplotlib.pyplot as plt
+from matplotlib import patches
+from matplotlib.axes import Axes
+
+from awpy.data.map_data import MAP_DATA
```
pnxenopoulos, line 17: This is a recent change, but I'm moving in favor of

```
import awpy.data.map_data
import awpy.plot.utils
import awpy.nav
```

then you can call them as `awpy.data.map_data.MAP_DATA` or `awpy.plot.utils.game_to_pixel` -- I tend to prefer this provenance a bit more.

awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
...
+            else:
+                edgecolor = "black"
+                facecolor = 'red'
+        else:
+            facecolor = "None"
+            edgecolor = "yellow"
+        _plot_tile(axis, polygon, edgecolor=edgecolor, facecolor=facecolor)
+
+def plot_map_tiles(map_name: str = "de_ancient", map_type: str = "original", output_path: str = "", 
+                   *, dark: bool = False, dpi: int = 1000, fill: str = "None") -> None:
```
pnxenopoulos, line 86: dpi = 300 is probably a better choice here, 1000 is very high

awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
...
+                facecolor = 'green'
+            else:
+                edgecolor = "black"
+                facecolor = 'red'
+        else:
+            facecolor = "None"
+            edgecolor = "yellow"
+        _plot_tile(axis, polygon, edgecolor=edgecolor, facecolor=facecolor)
+
+def plot_map_tiles(map_name: str = "de_ancient", map_type: str = "original", output_path: str = "",
```
pnxenopoulos, line 85: I don't think `map_type` is used here, is it?

awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    _plot_all_tiles(NAV[map_name], axis, default_fill=fill)
+    plt.show()
+    # Optionally save the figure:
+    if output_path != "":
+        plt.savefig(os.path.join(output_path, f"tiles_{map_name}.png"), bbox_inches="tight", dpi=dpi)
+    fig.clear()
+    plt.close()
+
+def plot_map_tiles_selected(map_name: str, selected_tiles: list, *,
+                            output_path: str = "", fill: str = "red", dpi: int = 1000) -> None:
```
pnxenopoulos, line 103: let's move in favor of `pathlib.Path` objects instead of `str` for paths, both here and anywhere else in this module. What we can do is say if `output` is equal to `None`, we don't write anything (`None` should be the default).

awpy/plot/tile_viz_utils.py
```diff
@@ -0,0 +1,128 @@
...
+import matplotlib.image as mpimg
+import matplotlib.pyplot as plt
+from matplotlib import patches
+from matplotlib.axes import Axes
+
+from awpy.data.map_data import MAP_DATA
+from awpy.plot.utils import game_to_pixel
+from awpy import Nav
+
+NAV = {}
```
pnxenopoulos, line 21: I'm going to come back to this, but leave it as is for now. Also, you can run ruff with `uv run ruff format .` and then `uv run ruff check --fix .`

awpy/plot/tile_viz_utils.py
adisujithkumar: Changed to `tile.py` in the next commit; thanks for the suggestion.

awpy/plot/tile_viz_utils.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/tile_viz_utils.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/tile_viz_utils.py
adisujithkumar: Ah thank you I forgot to clean this up; addressed in next commit.

awpy/plot/tile_viz_utils.py
adisujithkumar: Changed all default values for dpi to 300 in the next commit.

awpy/plot/tile_viz_utils.py
adisujithkumar: Addressed in the next commit to the branch; thanks for the recommendation!

adisujithkumar:
> I also have a PR with some api improvements for the nav module here: #364
> 
> Could probably propagate some changes over.

If it would be better to merge your nav changes first, then I totally can rebase my changes on top of yours no problem.

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+NAV["de_dust2"] = awpy.nav.Nav(path="../../awpy/data/de_dust2")
+
+
+def plot_map(map_name: str) -> tuple:
+    """Plots a map background based on the provided map name and returns the figure and axes objects.
+
+    This function loads a PNG image corresponding to the given map name and displays it using matplotlib.
+    If the map name ends with "_lower", that suffix is removed before loading the image.
+
+    Parameters:
```
pnxenopoulos, line 28: We use the Google docstring format. I'd just pass this into an LLM and ask it to change it do a Google-style Python docstring. You can see `awpy/demo.py` for an example of what a Google-style docstring looks like.

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Load and display the map background
+    map_bg = mpimg.imread(map_img_path)
+    ax.imshow(map_bg, zorder=0, alpha=0.5)
+
+    return fig, ax
+
+
+def _tile_polygon(area_dict: dict) -> list:
+    """Convert an area's corner coordinates to pixel coordinates."""
```
pnxenopoulos, line 69: For `_tile_polygon`, `_plot_tile`, and all other functions that are missing docstrings, I'd add them in the Google-style, even if they are `_internal` functions.

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+        print(f"The visualization has been saved at {save_path.resolve()}")
+    else:
+        plt.show()
+
+    fig.clear()
+    plt.close()
+
+
+def plot_map_tiles_selected(
+    map_name: str, selected_tiles: list, output_dir: Path | None = None, dpi: int = 300
```
pnxenopoulos, line 170: I wouldn't call it `output_dir` since we really want to specify a path here.

I'd consider something like `outpath: str | pathlib.Path | None = None`

Then, in the code, do
```
if outpath is not None:
    outpath = pathlib.Path(outpath)
```

That way people can give string or Path objects. This will still be safe is a Path object is passed.

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+    >>> plot_map_tiles_selected(
+    ...     map_name="de_dust2",
+    ...     selected_tiles=[5, 12, 18],
+    ...     output_dir="./maps",
+    ...     dpi=800
+    ... )
+    # Saves the plot to file at './maps/selected_tiles_de_dust2.png'
+    """
+    fig, axis = plot_map(map_name=map_name)
+    fig.set_size_inches(19.2, 21.6)
```
pnxenopoulos, line 214: How are `19.2` and `21.6` determined? Usually if I see a constant in code, it either means (1) it should be declared as a constant in a module, (2) it should become an argument to the method/function and passed by the caller.

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+    None
+        This function either displays or saves the plot if `output_dir` is provided.
+
+    Example:
+    --------
+    >>> plot_map_tiles(map_name="de_dust2", output_dir="./maps", dpi=800, fill="blue")
+    # Saves the plot to file at './maps/tiles_de_dust2.png'
+    """
+    fig, axis = plot_map(map_name=map_name)
+    fig.set_size_inches(19.2, 21.6)
```
pnxenopoulos, line 152: See later comment on fig size constants

awpy/plot/tile.py
```diff
@@ -0,0 +1,227 @@
...
+import awpy.nav
+import awpy.plot.utils
+
+NAV = {}
+MAP_DIR = "../data/nav"
+# TODO: Change hardcoded de_dust2
+NAV["de_dust2"] = awpy.nav.Nav(path="../../awpy/data/de_dust2")
+
+
+def plot_map(map_name: str) -> tuple:
```
pnxenopoulos, line 22: Just realized -- is this different than `plot(...)` in `awpy.plot.plot.py` ?

awpy/plot/tile.py
JanEricNitschke: Or you can use pymend (https://github.com/JanEricNitschke/pymend) :)

adisujithkumar: I've changed the docstring style in the next commit!

awpy/plot/tile.py
adisujithkumar: I've added docstrings for the internal functions in the next commit; feel free to comment on them regarding the tweaks you recommend.

awpy/plot/tile.py
adisujithkumar: I've made this change in the next commit!

awpy/plot/tile.py
adisujithkumar: Yeah I don't think we really need this size setting since in the old code it seems like we keep the figure size to be default? I can always add it as an argument if we want but for now I'll remove it.

awpy/plot/tile.py
adisujithkumar: Removed but can add back as an argument if we want

awpy/plot/tile.py
adisujithkumar: Nvm added the size setting back and explained further in the other comment

awpy/plot/tile.py
adisujithkumar: Lmao no it isn't; changed code to use `awpy.plot.plot`

awpy/plot/tile.py
adisujithkumar: Resolving this thread so we can use the other comment thread for sole conversation on figure size defaults

awpy/plot/tile.py
adisujithkumar: Created a [PR here](https://github.com/pnxenopoulos/awpy/pull/365) which moves the player plotting into a set of helper functions

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+import awpy.plot.utils
+
+NAV = {}
+MAP_DIR = "../data/nav"
+DEFAULT_FIG_SIZE = (19, 21)
+# TODO: Change hardcoded de_dust2 to contain data for all maps
+NAV["de_dust2"] = awpy.nav.Nav(path="../../awpy/data/de_dust2")
+
+
+def _tile_polygon(area_dict: dict, map_name: str) -> list:
```
pnxenopoulos, line 23: Could type the `dict` a bit more clearly

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+
+import matplotlib.pyplot as plt
+from matplotlib import patches
+from matplotlib.axes import Axes
+
+import awpy.nav
+import awpy.plot
+import awpy.plot.utils
+
+NAV = {}
```
pnxenopoulos, line 16: Need to remove these lines / check for the nav json file and read it that way

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+    Returns:
+        list: List of (x, y) pixel coordinates.
+
+    Example:
+        >>> _tile_polygon(area_dict, map_name="de_mirage")
+    """
+    return [awpy.plot.utils.game_to_pixel(map_name, (c["x"], c["y"], c["z"]))[0:2] for c in area_dict["corners"]]
+
+
+def _plot_tile(axis: Axes, polygon: list, edgecolor: str, facecolor: str, linewidth: int = 1) -> None:
```
pnxenopoulos, line 40: Typehint list to `list[tuple[float, float]]`

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+        >>> _plot_all_tiles(map_name, ax, default_fill="gray")
+    """
+    map_dict = NAV[map_name]
+    for area in map_dict.areas.values():
+        area_dict = area.to_dict()
+        polygon = _tile_polygon(area_dict, map_name)
+        _plot_tile(axis, polygon, edgecolor="yellow", facecolor=default_fill)
+
+
+def _plot_selected_tiles(map_name: str, axis: Axes, selected_tiles: list) -> None:
```
pnxenopoulos, line 81: typehint `selected_tiles` to `list[int]`

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+            edgecolor = "yellow"
+            facecolor = "None"
+        _plot_tile(axis, polygon, edgecolor=edgecolor, facecolor=facecolor)
+
+
+def plot_map_tiles(
+    map_name: str,
+    outpath: str | Path | None = None,
+    dpi: int = 300,
+    fill: str = "None",
```
pnxenopoulos, line 132: Instead of the string `"None"`, we should probably use the built-in `None`

awpy/plot/tile.py
```diff
@@ -0,0 +1,253 @@
...
+    figure_size: tuple[float, float] = DEFAULT_FIG_SIZE,
+) -> None:
```
pnxenopoulos, line 134: In these plotting functions, we should probably return the fig, ax -- this way we can also get rid of the set_size_inches (which we can use in the jupyter notebook)

awpy/plot/tile.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/tile.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/tile.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/tile.py
adisujithkumar: Addressed in next commit. Decided to just pass in the list of the NavArea's corners instead of the entire dict version of the NavArea. Let me know what you think

PR #364, opened 2025-02-20
Nav api improvements

JanEricNitschke:


pnxenopoulos:
I think all the nav changes look good. Two things:

(1) Since we don't use pyright in the linting process (hoping to move to whatever the Astral team packages into ruff), we don't need the pyproject.toml changes
(2) I updated the `.ps1` scripts, so we can use those.

JanEricNitschke:
Yeah, i just used pyright myself to check stuff and forgot to remove that again.

Will grab your updated scripts.

If there is no confusion where an import comes from i tend to prefer the shorter ones like `Path` and `Vector3` instead of having the full thing everywhere. If needed you can just look at the import line or use your ide to jump to the definition.

But in the end its your call and i can exapnd them back again.

PR #365, opened 2025-02-22
Moving Player Plotting To Helper Function(s) And Adding Datamodels To awpy.plot.plot

adisujithkumar:
@JanEricNitschke had a comment on [this PR](https://github.com/pnxenopoulos/awpy/pull/363) that it might be good to move player plotting functionality into a separate helper function instead of incorporating it into awpy.plot.plot. This PR handles that request. While working on it, I also generated a couple data models to use in the file as well - if this is overkill to introduce to awpy.plot lmk. Also if there are better/recommended ways to use the datamodels which Im not doing please let me know.

The testing I did for this change is running the `plot_demo.py` notebook and everything ran as expected on there.

### DIFF
--- a/awpy/plot/plot.py
+++ b/awpy/plot/plot.py
@@ -4,7 +4,8 @@ import importlib.resources
 import io
 import math
 import warnings
-from typing import Literal
+from dataclasses import dataclass
+from typing import Any, Literal, Optional
 
 import matplotlib.image as mpimg
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


awpy/plot/plot.py
```diff
@@ -77,161 +201,276 @@ def plot(  # noqa: PLR0915
...
+    return figure, axes
+
+def _plot_positions(
+    map_name: str,
+    axes: Axes,
+    points: list[tuple[float, float, float]] | None = None,
+    lower_points_frac: float | None = 0.4,
+    point_settings: list[PointSettings] | None = None,
+) -> None:
+    """Plots points on a map, optionally customizing plotting settings.
```
JanEricNitschke, line 218: Here it says we directly take a list of PointSettings but in the code you generate it from a dict. We should probably allow passing both in the signature and then do an instance check against dict and convert those?

awpy/plot/plot.py
```diff
@@ -20,7 +21,130 @@
...
+    Attributes:
+        x_pos (list[tuple[float, float]]): List of transformed x coordinates.
+        y_pos (list[tuple[float, float]]): List of transformed y coordinates.
+        plot_settings (list[PointSettings]): PointSettings for each plotted pair.
+    """
+
+    x_pos: list[tuple[float, float]]
+    y_pos: list[tuple[float, float]]
+    plot_settings: list[PointSettings]
+
```
JanEricNitschke, line 145: Instead of having a class of three lists that need to be the same length we could instead of a list of a class that has the three attributes?

awpy/plot/plot.py
```diff
@@ -77,161 +201,276 @@ def plot(  # noqa: PLR0915
...
 
-            transformed_x = awpy.plot.utils.game_to_pixel_axis(map_name, x, "x")
-            transformed_y = awpy.plot.utils.game_to_pixel_axis(map_name, y, "y")
+    return PlotPositionMetadata(x_pos=x_pos, y_pos=y_pos, plot_settings=updated_settings)
 
-            # Plot the marker
+def _plot_positions_helper(player_pos_settings: PlotPositionMetadata, axes: Axes) -> None:
+    """Plots player positions and associated metadata on a given matplotlib axes.
+
+    This function visualizes player positions on the map using the provided plot metadata.
```
JanEricNitschke, line 332: We can probably rename this from `helper` to something like `from_metadata`?

awpy/plot/plot.py
```diff
@@ -20,7 +21,130 @@
...
+    armor: Optional[int] = None
+    direction: Optional[tuple[float, float]] = None
+    label: Optional[str] = None
+    alpha: float = 1.0
+
+    def __post_init__(self) -> None:
+        """Validates the types and values of the attributes after initialization.
+
+        Ensures that each attribute matches its expected type and that values fall within
+        valid ranges. Raises appropriate exceptions if validation fails.
```
JanEricNitschke, line 55: Kind of rebuilding a pydantic model here? I feel we could either rely on a type checker to do this work instead of doing it at runtime or just using pydantic directly? Although it might not be worth that large dependency for just one case.

But i am actually leaning on just type checking. I think the only constraint that is not expressed in the types themselves is the alpha to being between 0 and 1?

awpy/plot/plot.py
adisujithkumar: Yup I agree. Thanks for the suggestion and addressed in next commit.

awpy/plot/plot.py
adisujithkumar: Yeah I agree this is a bit overkill. In the next commit I've changed this to just have type checking in the next commit like the other data classes in the package.

Since the alpha value check is a pretty minor validation (imo), I'm just going to ignore it for now. If you think it would be good to include validation for that, I can introduce that later. Keeping this conversation unresolved till we decide on whether alpha attribute validation is needed.

awpy/plot/plot.py
adisujithkumar: Yeah I agree it would be good to handle both in the signature. I tried addressing this in the next commit; keeping this conversation as unresolved since you might still have some comments on my changes

awpy/plot/plot.py
```diff
@@ -213,7 +154,7 @@ def _plot_positions(
     axes: Axes,
     points: list[tuple[float, float, float]] | None = None,
     lower_points_frac: float | None = 0.4,
-    point_settings: list[PointSettings] | None = None,
+    point_settings: list[PointSettings] | list[dict] | None = None,
 ) -> None:
```
JanEricNitschke: I would also use `dict[str, Any]` here like in  the signature of `from_dict`

awpy/plot/plot.py
```diff
@@ -260,20 +201,24 @@ def _plot_positions(
         settings_mismatch_err = "Number of points and point_settings do not match."
         raise ValueError(settings_mismatch_err)
     else:
-        point_settings = [PointSettings.from_dict(setting) for setting in point_settings]
+        # If dicts are passed into the function
+        # convert them to PointSettings objects
+        for i in range(len(point_settings)):
+            if type(point_settings[i]) is dict:
+                point_settings[i] = PointSettings.from_dict(point_settings[i])
```
JanEricNitschke, line 210: `isinstance` instead of `type(...) is`

awpy/plot/plot.py
```diff
@@ -319,14 +262,15 @@ def _generate_plot_metadata(
...
-        y_pos.append(transformed_y)
-        updated_settings.append(updated_point)
+        # Store transformed coordinates and updated settings in PlotPositionMetadata obj
+        # and add it to the list
+        plot_position_metadata_list.append(PlotPositionMetadata(x_pos=transformed_x,
+                                                          y_pos=transformed_y,
+                                                          plot_settings=updated_point_settings))
 
-    return PlotPositionMetadata(x_pos=x_pos, y_pos=y_pos, plot_settings=updated_settings)
+    return plot_position_metadata_list
```
JanEricNitschke: Is there no CI running here? Because i think the formatter should complain about this.

awpy/plot/plot.py
```diff
@@ -260,20 +201,24 @@ def _plot_positions(
...
     point_settings: list[PointSettings],
     lower_points_frac: float = 0.4,
-) -> PlotPositionMetadata:
+) -> list[PlotPositionMetadata]:
     """Processes points and their settings to prepare plotting metadata.
 
-    Args:
+    Args:0
         map_name (str): Name of the map.
         points (list[tuple[float, float, float]]): List of (x, y, z) points.
```
JanEricNitschke: I think a `0` slipped in here.

awpy/plot/plot.py
JanEricNitschke: Nah, i thinks its fine to leave that out.

awpy/plot/plot.py
JanEricNitschke: Added some new comments. Almost where i think it should be

awpy/plot/plot.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/plot.py
adisujithkumar: Addressed in next commit; thanks!

awpy/plot/plot.py
adisujithkumar: Doesn't seem to be there anymore so I must have cleaned it up last night

awpy/plot/plot.py
adisujithkumar: CI as in ruff? I ran the two ruff commands through uv (`uv run ruff format .` and `uv run ruff check --fix`) and didn't get any complaints. Is there something else I should run?

JanEricNitschke: Weird, because i think ruff doesnt format things like this:

```
        plot_position_metadata_list.append(PlotPositionMetadata(x_pos=transformed_x,
                                                          y_pos=transformed_y,
                                                          plot_settings=updated_point_settings))
```

https://play.ruff.rs/b33054c5-3414-45c6-8c27-565754db07fd

JanEricNitschke:
Looks good to me now.

awpy/plot/plot.py
adisujithkumar: I changed it to that since the line was too long based on character length.

Does the ruff check fail for you on the branch? Just wondering if my ruff is set up wrong or something.

PR #379, opened 2025-03-06
fix round end official ticks

pnxenopoulos:
As pointed out by btrams: "by default the parser used to filter out freeze time ticks and consider ticks between end and official_end, right now the behaviour's reversed"

### DIFF
--- a/awpy/parsers/rounds.py
+++ b/awpy/parsers/rounds.py
@@ -66,7 +66,7 @@ def _add_bomb_plant_info(rounds_df: pl.DataFrame, bomb_plants: pl.DataFrame) -> 
     For each round, this function looks for bomb plant events occurring between
     the round's start and end ticks. It then adds two new columns:
       - bomb_plant: The tick at which the bomb was planted (if any).
-      - bomb_site: "bombsite a" or "bombsite b" based on the site value, or "not planted"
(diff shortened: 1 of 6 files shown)


JanEricNitschke:
What was "end" and "official_end" again? end is when one side won and official_end when it switches to the next round


Otherwise seems to look good to me

awpy/parsers/rounds.py
```diff
@@ -233,7 +233,7 @@ def apply_round_num(df: pl.DataFrame, rounds_df: pl.DataFrame, tick_col: str = "
     # Validate that the event tick is within the round boundaries.
     # If the tick is greater than the round's 'end', then set round_num to null.
     df_with_round = df_with_round.with_columns(
-        pl.when(pl.col(tick_col) <= pl.col("end")).then(pl.col("round_num")).otherwise(None).alias("round_num")
+        pl.when(pl.col(tick_col) < pl.col("official_end")).then(pl.col("round_num")).otherwise(None).alias("round_num")
```
smartbackwards, line 236: please revert to <= to avoid not parsing the last event of the game

pnxenopoulos: Nice catch, done

PR #384, opened 2025-03-15
Fix bullet point numbering in demo.py

n1c:


pnxenopoulos:
Perfect, thank you for this! Feel free to open doc PRs whenever, even if it is to include an example that you think will be helpful.

Issue #395, opened 2025-04-10
Clarify how to pass events to parse (and add a test for it)


JanEricNitschke:
Currently i feel it is not super clear how i should pass events, player_props etc on the CLI if i want to manually specify the ones that i want parsed.

Also couldnt find a test for that.

Code looks a bit weird to me in that it should probably be passed as 

`awpy parse DEMO_PATH --events="EVENT1,EVENT2,..."`? Which i dont think is what i would usually expect.

PR #397, opened 2025-04-11
Improve handling when specific events are missing.

### LINKED ISSUE #396: CLI `awpy parse` crashes w/ `KeyError: "player_sound"` on default settings for some demos
**Describe the bug**
I ran `awpy parse DEMO_PATH` and got a `KeyError` saying that `player_sound` isn't present. This seems to be a common issue with many demos, and hard stops the `parse` command. Since `player_sound` is often not included, and is parsed by default in `awpy parse`, it would be good to handle missing properties with a warning.

There are 2 obvious solutions off the top of my head.

First, we could remove `player_sound` from the default set of events to parse. This might be correct to do anyway, but I think that even when we specify a property that isn't in the demo awpy cli should handle it gracefully (either exit with a nice error message or continue optimistically).

Second, we can wrap our cachedproperties in a try/except:

```python


    @cached_property
    def footsteps(self) -> pl.DataFrame:
        try:
            footsteps = awpy.parsers.utils.get_event_from_parsed_events(self.events, "player_sound")
            footsteps = awpy.parsers.events.parse_footsteps(footsteps)
            return awpy.parsers.rounds.apply_round_num(df=footsteps, rounds_df=self.rounds, tick_col="tick").filter(
                pl.col("round_num").is_not_null()
            )
        except KeyError as e:
            print_pretty_user_warning("footsteps", e)
            return pl.DataFrame()
```

This second method is also a bit awkward because we have an empty dataframe, so maybe we can use optional types instead for the cached properties?

**Include Information to Reproduce**
Run `awpy parse DEMO` for a valve/faceit demo (i seems like it works on some pro demos, I don't know if it has 100% failure rate on valve/faceit demos, but it's common enough that it should fail after 3 or 4 tries at most).

This produces the following stack trace:

```
awpy parse /mnt/c/Program\ Files\ \(x86\)/Steam/steamapps/common/Counter-Strike\ Global\ Offensive/game/csgo/replays/match730_003740035103774998617_0611549455_129.dem
2025-04-10 11:51:01.073 | WARNING  | awpy.data.map_data:<module>:20 - Failed to load map data from /home/benku/.awpy/maps/map-data.json.
Traceback (most recent call last):
  File "/home/benku/Code/awpy/.venv/bin/awpy", line 8, in <module>
    sys.exit(awpy_cli())
             ^^^^^^^^^^
  File "/home/benku/Code/awpy/.venv/lib/python3.11/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/.venv/lib/python3.11/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/.venv/lib/python3.11/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/.venv/lib/python3.11/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/.venv/lib/python3.11/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/awpy/cli.py", line 73, in parse_demo
    demo.compress(outpath=outpath)
  File "/home/benku/Code/awpy/awpy/demo.py", line 670, in compress
    ("footsteps", self.footsteps),
                  ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/awpy/demo.py", line 375, in footsteps
    footsteps = awpy.parsers.utils.get_event_from_parsed_events(self.events, "player_sound")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benku/Code/awpy/awpy/parsers/utils.py", line 27, in get_event_from_parsed_events
    raise KeyError(missing_key_err_msg)
KeyError: "Required event 'player_sound' is missing from the events dictionary."
```

JanEricNitschke:
Closes: https://github.com/pnxenopoulos/awpy/issues/396

### DIFF
--- a/awpy/demo.py
+++ b/awpy/demo.py
@@ -499,8 +499,6 @@ class Demo:
             list[str]: A list of default event names.
         """
         return [
-            "round_freeze_end",  # Freeze time ends
-            "round_officially_ended",  # Round officially declared over
             "player_spawn",  # Players spawn
             "player_given_c4",  # C4 is given to a player
             "bomb_pickup",  # Bomb is picked up
@@ -566,6 +564,8 @@ class Demo:
         # Explicitly parse round start and round end events
         events["round_start"] = self.parser.parse_event("round_start")
         events["round_end"] = self.parser.parse_event("round_end")
+        events["round_freeze_end"] = self.parser.parse_event("round_freeze_end")
+        events["round_officially_ended"] = self.parser.parse_event("round_officially_ended")
 
         # Loop through and process each event
         for event_name, event in events.items():
--- a/awpy/demo.py
+++ b/awpy/demo.py
@@ -372,7 +372,9 @@ class Demo:
             KeyError: If 'player_sound' events are not found in the parsed events. This may
                     indicate that the demo has not been parsed yet. Please run the .parse() method.
         """
-        footsteps = awpy.parsers.utils.get_event_from_parsed_events(self.events, "player_sound")
+        footsteps = awpy.parsers.utils.get_event_from_parsed_events(
+            self.events, "player_sound", empty_if_not_found=True
+        )
         footsteps = awpy.parsers.events.parse_footsteps(footsteps)
         return awpy.parsers.rounds.apply_round_num(df=footsteps, rounds_df=self.rounds, tick_col="tick").filter(
             pl.col("round_num").is_not_null()
@@ -535,7 +537,7 @@ class Demo:
 
     def parse_events(
         self,
-        events_to_parse: list[str] | None = None,
+        events_to_parse: list[str],
         player_props: list[str] | None = None,
         other_props: list[str] | None = None,
     ) -> dict[str, pl.DataFrame]:
--- a/awpy/demo.py
+++ b/awpy/demo.py
@@ -372,9 +372,7 @@ class Demo:
             KeyError: If 'player_sound' events are not found in the parsed events. This may
                     indicate that the demo has not been parsed yet. Please run the .parse() method.
         """
-        footsteps = awpy.parsers.utils.get_event_from_parsed_events(
-            self.events, "player_sound", empty_if_not_found=True
-        )
+        footsteps = awpy.parsers.utils.get_event_from_parsed_events(self.events, "player_sound")
         footsteps = awpy.parsers.events.parse_footsteps(footsteps)
         return awpy.parsers.rounds.apply_round_num(df=footsteps, rounds_df=self.rounds, tick_col="tick").filter(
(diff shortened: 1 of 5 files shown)


JanEricNitschke:
Also improved the args handling for `parse` a bit.

Closes: https://github.com/pnxenopoulos/awpy/issues/395

awpy/demo.py
```diff
@@ -664,18 +664,23 @@ def compress(self, outpath: Path | None = None) -> None:
...
+                "shots",
+                "grenades",
+                "infernos",
+                "smokes",
+                "bomb",
+                "ticks",
+                "rounds",
             ]:
+                try:
+                    parsed_df: pl.DataFrame = getattr(self, df_name)
```
bkushigian, line 680: yeah this is a lot cleaner IMO

awpy/parsers/utils.py
```diff
@@ -1,6 +1,7 @@
 """Module for parsing utils."""
 
 import polars as pl
+from loguru import logger
 
 
 def get_event_from_parsed_events(
```
bkushigian, line 7: I would argue this should be `Optional[pl.DataFrame]` so that the type is large enough to have all possible outcomes (no list + empty list). But this might not be an issue in practice and would require some extra work elsewhere in the code. Plus, optionals are only good if their usage is enforced by the type system, so this is probably better.

bkushigian (commented):
LGTM. I have one comment about the type of `utils::get_event_from_parsed_events` but I'm not convinced, just putting it up there for consideration.

n1c:
Is it on the cards for this to get merged? Bumped into an issue today that I think this will solve.

